PR #3, opened 2017-04-04
<test>: Add test for latency

mbulatova:
For performance tracking purpose test for packet latency is needed
in addition to tests for throughput. This test measures time
between sending a packet and getting a response to it.

Signed-off-by: Maria Bulatova <maria.bulatova@intel.com>

### DIFF
--- a/test/main/latency.json
+++ b/test/main/latency.json
@@ -1,0 +1,86 @@
+{
+    "docker-config": {
+        "request-timeout": 10000000000,
+        "docker-client-version": "1.24",
+        "privileged": true,
+        "map-volumes": [
+            "/sys/bus/pci/drivers:/sys/bus/pci/drivers",
+            "/sys/kernel/mm/hugepages:/sys/kernel/mm/hugepages",
+            "/sys/devices/system/node:/sys/devices/system/node",
+            "/dev:/dev"
+        ],
+        "docker-port": 2375,
+        "pktgen-port": 22022
+    },
+    "tests": [
+        {
+            "name": "latency-test-light",
+            "test-time": 60000000000,
+            "test-type": "TEST_TYPE_BENCHMARK",
+            "test-apps": [
+                {
+                    "host-name": "hostname1",
+                    "image-name": "yanff-performance",
+                    "app-type": "TESTAPP_GO",
+                    "exec-cmd": [
+                        "./latency-part1"
+                    ]
+                },
+                {
+                    "host-name": "hostname2",
+                    "image-name": "yanff-performance",
+                    "app-type": "TESTAPP_GO",
+                    "exec-cmd": [
+                        "./perf_light"
+                    ]
+                }
+            ]
+        },
+		{
+            "name": "latency-test-seq",
+            "test-time": 60000000000,
+            "test-type": "TEST_TYPE_BENCHMARK",
+            "test-apps": [
+                {
+                    "host-name": "hostname1",
+                    "image-name": "yanff-performance",
+                    "app-type": "TESTAPP_GO",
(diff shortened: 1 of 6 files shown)


test/latency/latency.json
```diff
@@ -0,0 +1,40 @@
...
+        "pktgen-port": 22022
+    },
+    "tests": [
+        {
+            "name": "latency-test",
+            "test-time": 60000000000,
+            "test-type": "TEST_TYPE_SCENARIO",
+            "test-apps": [
+                {
+                    "host-name": "nonpcmp2",
```
gshimansky: Let's use generic hostname1 and hostname2 in test configs because that is what we have in documentation at the moment.

test/latency/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# Copyright 2017 Intel Corporation. 
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+FROM nfvbase
+
```
gshimansky: nfvbase has been renamed to yanff-base

test/latency/latency-part1.go
```diff
@@ -0,0 +1,198 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"fmt"
+	"nfv/flow"
+	"nfv/packet"
+	"sync"
```
gshimansky: Import paths should be corrected to github.com/intel-go/yanff/

gshimansky:
In addition to the above comments, please commit only Unix line endings files. In the current patch I see CRLFs in Dockerfile, .gitignore and latency.json.

test/latency/latency-part1.go
```diff
@@ -0,0 +1,198 @@
...
+	testDoneEvent.L.Lock()
+	testDoneEvent.Wait()
+	testDoneEvent.L.Unlock()
+
+	// Compose statistics
+	sent := atomic.LoadUint64(&sentPackets)
+	checked := atomic.LoadUint64(&checkedPackets)
+
+	// Leave time for goroutine to print latencies info to screen
+	time.Sleep(time.Second)
```
gshimansky: Using Sleep is bad design. It is much better to use events or channels for such purpose. You are using testDoneEvent a few lines above. Sending such event could be done in line 128 of latenciesLogger.

test/latency/latency-part1.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// Event needed to stop sending packets
+	var m sync.Mutex
+	testDoneEvent = sync.NewCond(&m)
+
+	// Initialize YANFF library at 16 available cores
+	flow.SystemInit(16)
+
+	// Create packet flow
+	outputFlow := flow.SetGenerator(generatePackets, 10000)
```
gshimansky: It is better to specify this constant at the beginning of the program.

test/latency/latency-part1.go
```diff
@@ -0,0 +1,198 @@
...
+	flow.SetSender(outputFlow, 0)
+
+	// Create receiving flow and set a checking function for it
+	inputFlow := flow.SetReceiver(1)
+
+	// Calculate latency only for 1 of 100000 packets. Otherwise channel is conjested
+	latFlow := flow.SetPartitioner(inputFlow, 100000, 1)
```
gshimansky: It is better to specify this constant at the beginning of the program.

mbulatova:
Fixed according to remarks. Added cmd line flags to set parameters for this test (packet size, generator speed, etc.) Tested with test system.

PR #4, opened 2017-04-11
Added target "make testing" to run unit tests

gshimansky:
New target for unit tests is separate from just building everything because to run tests it is necessary to become root while compilation may be done by a normal user.


ifilippov:
Looks good to me

PR #5, opened 2017-04-14
New functions for packet package

ifilippov:
Added Encapsulation and Decapsulation functions for packet:
EncapsulateHead, EncapsulateTail, DecapsulateHead, DecapsulateTail
which add or remove bytes inside a packet.

Added SwapBytesUint32, GetPacketLen, PacketBytesChange functions.
Made GetRawBytes function a member of packet package.
Made small type corrections.
Corresponding changes in low package and tests, examples.

### DIFF
--- a/packet/packet.go
+++ b/packet/packet.go
@@ -554,7 +554,7 @@ func (packet *Packet) CreatePacket(IN uintptr) {
 // PacketFromByte function gets non-initialized packet and slice of bytes of any size.
 // Initializes input packet and fills it with these bytes.
 func PacketFromByte(packet *Packet, data []byte) {
-	low.AppendMbuf(packet.CMbuf, len(data))
+	low.AppendMbuf(packet.CMbuf, uint(len(data)))
 	low.WriteDataToMbuf(packet.CMbuf, data)
 	packet.CreatePacket(uintptr(unsafe.Pointer(packet.CMbuf)))
 }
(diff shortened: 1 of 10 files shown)


test/stash/create_packet_example.go
```diff
@@ -38,8 +38,7 @@ func main() {
 
 func generatePacket(pkt *packet.Packet) {
 	// Total packet size will be 14+20+20+70+4(crc)=128 bytes
-	size := 70
-	packet.InitEmptyEtherIPv4TCPPacket(pkt, size)
+	packet.InitEmptyEtherIPv4TCPPacket(pkt, 70)
 	pkt.Ether.DAddr = [6]uint8{0xde, 0xad, 0xbe, 0xaf, 0xff, 0xfe}
```
gshimansky: Change 0xaf to 0xef :)

packet/packet.go
```diff
@@ -686,11 +686,91 @@ func InitEmptyEtherIPv6UDPPacket(packet *Packet, plSize int) {
...
+
+// EncapsulateTail adds bytes to packet. start - number of beginning byte, length - number of
+// added bytes. This function should be used to add bytes to the second half
+// of packet. Return false if error.
+func (packet *Packet) EncapsulateTail(start uint, length uint) bool {
+	if low.AppendMbuf(packet.CMbuf, length) == false {
+		return false
+	}
+	packetLength := packet.GetPacketLen()
+	for i := packetLength - 1; int(i)-int(length) >= int(start); i-- {
```
gshimansky: Possibly remove condition calculation by changing loop condition to int(i) >= int(start)+int(length) because int(start)+int(length) may be calculated just once while int(i)-int(length) has to be calculated on every loop iteration.

PR #7, opened 2017-04-20
Add more control of the system parameters for developers

mbulatova:
Before all default values of system parameters were set in code
and then modified via cmd line options. This commit adds more control:
1. New default values can be specified by developer with JSON-formatted
string which is passed to SystemInit().
2. Default values can be locked from user change.

Flags creation and default values assignment move to common package.
Parameter CPUCoresNumber is made a field of struct OptsSettings.
Move debug print of options to package common. Update examples.

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -30,7 +30,6 @@
 package flow
 
 import (
-	"flag"
 	"github.com/intel-go/yanff/asm"
 	"github.com/intel-go/yanff/common"
 	"github.com/intel-go/yanff/low"
@@ -201,11 +200,6 @@ func makeHandler(in *low.Queue, out *low.Queue,
 	return scheduler.NewClonableFlowFunction("handler", ffCount, handle, par, handleCheck, make(chan uint64, 50))
 }
 
-var burstSize uint
-var sizeMultiplier uint
-var schedTime uint
-var maxPacketsToClone uint32
-
 type port struct {
 	rxQueues       []bool
 	txQueues       []bool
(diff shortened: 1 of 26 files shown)


flow/flow.go
```diff
@@ -224,24 +218,64 @@ const (
...
+//  var optsExample string =
+//  `{"mempool-size":     {"Value": 8191,  "Locked": true},
+//    "no-remove-clones": {"Value": false, "Locked": true},
+//    "ring-size":        {"Value": 512,   "Locked": true},
+//    "burst-size":       {"Value": 64,    "Locked": false}}`
+// In example string mempool-size and no-remove-clones default values are not changed,
+// but locked. Ring-size default value is modified (was 256) and locked.
+// Burst-size default value is modified (was 32) and not locked.
+//
+// Available options to modify in the config string:
```
ifilippov: Available options with comments can be seen in common/common.go file

flow/flow.go
```diff
@@ -224,24 +218,64 @@ const (
...
+//   cores
+//   no-scheduler
+//   no-remove-clones
+//   mempool-size
+//   mempool-cache
+//   ring-size
+//   scale-time
+//   burst-size
+// By default, parameters which are absent in config string are not locked.
```
ifilippov: and have values defined in common/common.go file.

ifilippov:
Four small changes in comments and looks good to me.

gshimansky:
Why did you decide to use JSON at all? I haven't seen JSON used in any Go library. Take a look at [docker library](https://godoc.org/github.com/moby/moby/client#Client.ContainerCreate). Functions that require many parameters take them inside of big structures. When parameter it not present, it is initialized as zero/false. If you absolutely need to have a flexible parameter structure, you could use a "map[string]int" type where you could specify optional parameters. If number of cores is a mandatory parameter, it should b present in function prototype.

ifilippov:
I disagree because it is a way in which parameters is passed in DPDK.
	struct rte_eth_conf port_conf_default = {
	    .rxmode = { .max_rx_pkt_len = ETHER_MAX_LEN,
			.mq_mode = ETH_MQ_RX_RSS    },
	    .txmode = { .mq_mode = ETH_MQ_TX_NONE, },
	    .rx_adv_conf.rss_conf.rss_key = NULL,
	    .rx_adv_conf.rss_conf.rss_hf = ETH_RSS_PROTO_MASK
	};
I suppose that JSON is not a bad variant in this situation. Are you completely against it?

gshimansky:
Well, I very much dislike this approach. It is very uncommon, in fact I haven't seen it used in any other libraries. The most common approach would be to make a set of functions, each of them would allow to specify one or several parameters. This is how it is done in most cases. Using a big structure is Ok too, in fact it is a Go way of doing such stuff unless you don't like the default value for uninitialized structure field (e.g. zero is not a good default value).
They way DPDK initialization is done is an example of an ugly bicycle invention where it is not required.

aregm:
I think Grisha is right - it looks bad, it is nonstandard and developers will hate it. With the right given to me by <typename Authority> vetoing it. Let's come up with more standard and elegant way.

PR #8, opened 2017-05-01
combining cores for stop and scheduler

ifilippov:
Normally stop shouldn't be used at all time. It should be used for dropping wrong packets and dropping forbidden packets. It is strange situation when user wants to drop half of input flow, however we left a specific option for this situation.

Now stop is used more because we have problems with rings. However it is not normal situation and after this we are in recovery mode. So in normal conditions we shouldn't support this.

Performance testing is in progress. No merge before it will be finished.


gshimansky:
Please change the name of the flag to something like RunStopOnDedicatedCPUCore. Don't use word "separate" because we have a Separate flow function. After reading your comment and code I was under an impression that you want to make a combined function Separate+Stop flow function.

ifilippov:
Fixed

PR #9, opened 2017-05-08
Improving scheduler

ifilippov:
Adding a slice of speeds for each flow function for each
number of clones instead of one previous value.
Both cloning and removing clones update these speeds before changing anything
Removing clones works if current speed is less than saved speed with decreased number of clones (not changed)
Adding clones works if current speed is less that saved speed with increased number of clones
or we dont know speed with increased number of clones (new)
After some time period going from user system "forgets" speed with increased number of clones
and checks it by adding one cloning. It is added for changing flow functions behaviour

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -18,9 +18,9 @@ import (
 // expect this.
 var ScaleTime uint
 
-// TODO: 0.98 is a delta of speeds. This delta should be
+// TODO: 1 is a delta of speeds. This delta should be
 // verified or should be tunable from outside.
-const speedDelta = 0.98
+const speedDelta = 1
 
 // Clones of flow functions. They are determined
 // by their core and their stopper channel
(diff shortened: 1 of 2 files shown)


ifilippov:
Ping?))

scheduler/scheduler.go
```diff
@@ -147,11 +152,22 @@ func (scheduler *Scheduler) SystemStart() {
...
 	for {
 		time.Sleep(time.Millisecond * time.Duration(schedTime))
 		common.LogDebug(common.Debug, "System is using", scheduler.usedCores, "cores now.", uint8(len(scheduler.freeCores))-scheduler.usedCores, "cores are left available.")
 		for i := range scheduler.Clonable {
 			ff := scheduler.Clonable[i]
+			select {
+			// After checkTime milliseconds we suppose that flow function behaviour changed and want
+			// to try to add some clones (if can). If adding clones decrease performance we will
+			// stop them and wait for another checkTime period. If we know speed with more clones
+			// we forget it.
```
gshimansky: Regular flow function speed check after each checkTime milliseconds, it is done by erasing previously measured speed data. If it is possible, scheduler adds some new clones. If performance decreases with more added clones, they are stopped until another checkTime milliseconds pass.

ifilippov:
Updated with more structural comments.

PR #11, opened 2017-05-16
Add transfer settings to YANFF library from user app as struct

mbulatova:
Before values of system parameters were set in code and
then modified via cmd line options. Now settings are passed
as a struct to SystemInit() function. Update examples and tests.

### DIFF
--- a/examples/Firewall.go
+++ b/examples/Firewall.go
@@ -12,13 +12,16 @@ import (
 )
 
 var L3Rules *rules.L3Rules
+var cores uint
 
 // Main function for constructing packet processing graph.
 func main() {
-	// Initialize YANFF library at requested number of cores
-	var cores uint
 	flag.UintVar(&cores, "cores", 8, "Number of cores to use by system")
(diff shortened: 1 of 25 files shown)


aregm:
Wait - let's discuss the design. And some implementation questions.

mbulatova:
This motivation is to not oblige the user to use the command line to transfer parameters. If user want to modify settings via cmd line options - he is free to do so (please, see for example 'nosched' option in  test/performance/perf_light.go).

mbulatova:
Please, do not merge. I see one problem: if user want to create option for cores' number, which is parameter of SystemInit, this option will not be parsed before SystemInit call and as result SystemInit is always called with default 'cores' value. This is due to parsing is done too late, only inside SystemInit.
For this case user need to call flag.Parse() before SystemInit. But now we prohibit this, because this will prevent other flags (located in low.go) to be added.

mbulatova:
This change was revised by team and rejected.

PR #13, opened 2017-05-26
Change user-defined functions API

ifilippov:
Two main improvements which changed usir API

    1) Added vector variants of user-defined fucntions:
    vectorHandleFunction, vectorSeparateFunction, vectorGenerateFunction.
    These functions should help to achieve more performance by providing
    burstSize number of packets to user at one time.

    2) Added context parameter to all user defined functions and
    its definition in Set functions. Context should be a struct with
    Copy method. All not-thread-safe global variables should be
    inside this structure to be copied when scaling. Now all contextes are nil.
    Real contextes will be added shortly.

### DIFF
--- a/packet/packet.go
+++ b/packet/packet.go
@@ -57,6 +57,7 @@ const (
 
 // Supported L4 types
 const (
+	IPNumber  = 0x04
 	TCPNumber = 0x06
 	UDPNumber = 0x11
 )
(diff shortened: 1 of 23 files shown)


packet/packet.go
```diff
@@ -551,6 +552,12 @@ func (packet *Packet) CreatePacket(IN uintptr) {
 	packet.CMbuf = mbuf
 }
 
+func CreatePackets(packet []*Packet, IN []uintptr, n uint) {
```
gshimansky: Please add godoc comment about this function. What it does and how.

scheduler/scheduler.go
```diff
@@ -29,9 +29,13 @@ type clonePair struct {
 	channel chan int
 }
 
+type Context interface {
```
gshimansky: The name could be confused with context.Context which is a widely used Go type. If possible, we should find another name for this struct.

PR #19, opened 2017-06-14
Updates to build system to check that DPDK is built successfully

gshimansky:
Moved DPDK to top level because its location in tests was confusing


ifilippov:
Looks good to me

PR #20, opened 2017-06-15
Fixed pktgen build because in parallel it fails sometimes

gshimansky:


ifilippov:
Looks good to me. Maybe we need to ask to fix this in pktgen?

gshimansky:
Keith wrote that parallel build is not supported by pktgen https://github.com/pktgen/Pktgen-DPDK/issues/102

PR #21, opened 2017-06-16
Storing packet structure at the beginning of the mbuf structure

ifilippov:
YANFF has a limitation that parsed packets are loosen after exiting flow function.
We can't save packet structure (8 pointers to protocol headers) anywhere for the
following flow function because we need some memory allocation which is slow.

However each mbuf have "headroom" before packet payload. This "headroom" is used
for fast prepending to payload, for example while encapsulation. The length of
headroom is a property of DPDK and can be changed in dpdk/config/common_base file
in CONFIG_RTE_PKTMBUF_HEADROOM. We can use some of this memory for storing
packet structure there. For this we should:

1) Clean and prepare this headroom after receiving packet
   and after allocating mbuf for generation. This required
   removing "-g" from C compiling and using of fast AVX instructions.
2) Consider size of packet structure in prepend function
3) Change CreatePacket functions that creates packets from mbuf
   to ExtractPacket functions which extracts this memory from mbuf.
   This requires changing of prefetchs, they need to prefetch
   different line.

Stability was checked, performance of no-scheduler increased for 4%, number of cores after switching scheduler on decreased for 3%.

### DIFF
--- a/low/low.c
+++ b/low/low.c
@@ -17,11 +17,36 @@
 
 // #define DEBUG
 // #define VERBOSE
+#define AVX
 
-struct rte_mempool * mp;
+#ifdef AVX
+#define mbufClear(buf) \
+	_mm256_storeu_ps((float*)(buf + mbufStructSize), zero256);
+#else
+#define mbufClear(buf) \
+	_mm_storeu_ps((float*)(buf + mbufStructSize), zero128); \
+	_mm_storeu_ps((float*)(buf + mbufStructSize + 16), zero128);
+#endif
+
+#define mbufInit(buf) \
+	mbufClear(buf) \
+	_mm_storeu_ps((float*)(buf + mbufStructSize + 32), zero128); \
+	*(char**)((char*)(buf) + mbufStructSize + 48) = (char*)(buf) + defaultStart; \
+	*(char**)((char*)(buf) + mbufStructSize + 56) = (char*)(buf);
+
 long receive_received = 0, receive_pushed = 0;
 long send_required = 0, send_sent = 0;
 long stop_freed = 0;
+
+int mbufStructSize;
+int headroomSize;
+int defaultStart;
+
+__m128 zero128 = {0, 0, 0, 0};
+#ifdef AVX
+__m256 zero256 = {0, 0, 0, 0, 0, 0, 0, 0};
+#endif
+
 // This is multiplier for conversion from PKT/s to Mbits/s for 64 (84 in total length) packets
 const float multiplier = 84.0 * 8.0 / 1000.0 / 1000.0;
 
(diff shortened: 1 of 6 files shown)


low/low.c
```diff
@@ -17,11 +17,34 @@
...
 long send_required = 0, send_sent = 0;
 long stop_freed = 0;
+
+int mbufStructSize;
+int headroomSize;
+int defaultStart;
+
+__m128 zero128;
+__m256 zero256;
+
```
gshimansky: Add initialization with zeroes to these constants and a const modifier if compiler allows it for C code.

low/low.c
```diff
@@ -221,6 +250,14 @@ void eal_init(int argc, char *argv[], uint32_t burstSize)
 	free(argv[argc-1]);
 	free(argv);
 	BURST_SIZE = burstSize;
+	mbufStructSize = sizeof(struct rte_mbuf);
+	headroomSize = RTE_PKTMBUF_HEADROOM;
+	defaultStart = mbufStructSize + headroomSize;
+	zero128 = _mm_xor_ps(zero128, zero128);
+#ifdef AVX
+	zero256 = _mm256_xor_ps(zero256, zero256);
```
gshimansky: If compiler allows initialization at variables declaration, these XORs are not needed.

low/low.c
```diff
@@ -108,14 +132,19 @@ void recv(uint8_t port, uint16_t queue, struct rte_ring *out_ring, uint8_t coreI
 		if (unlikely(rx_pkts_number == 0))
 			continue;
 
+		for (i = 0; i < rx_pkts_number; i++) {
+			// TODO refetcht here
+			mbufInit(bufs[i]);
```
gshimansky: Make sure that only stores happen inside of the loop, loads or xors are outside of the loop.

low/low.go
```diff
@@ -52,10 +53,24 @@ func GetPacketDataStartPointer(mb *Mbuf) uintptr {
...
 }
 
+var packetStructSize int
+
+func SetPacketStructSize(t int) {
+	if t > C.RTE_PKTMBUF_HEADROOM {
+		common.LogError(common.Initialization, "Packet structure can't be placed inside mbuf.",
+			"Increase CONFIG_RTE_PKTMBUF_HEADROOM in dpdk/config/common_base and rebuild dpdk.")
+	}
+	if C.RTE_PKTMBUF_HEADROOM-t < 64 {
```
gshimansky: Change 64 to a named constant to add meaning to it

low/low.c
ifilippov: Checked, no differences in performance for xor, load or init inside function.

PR #23, opened 2017-06-19
Fixed parallel build to generate necessary files

gshimansky:
Fixed go generate which didn't work after parallel build changes


ifilippov:
Please, commit

PR #24, opened 2017-06-20
Conditional selection of AVX/AVX2 code based on machine CPU

gshimansky:


low/low.c
```diff
@@ -17,9 +17,8 @@
 
 // #define DEBUG
 // #define VERBOSE
-#define AVX
 
-#ifdef AVX
+#ifdef RTE_MACHINE_CPUFLAG_AVX2
```
ifilippov: Change for AVX

TocarIP: AVX2 is not needed, AVX is enough

PR #25, opened 2017-06-21
Patch DPDK source for VM.

sbilly:
Patch DPDK source for VM. from: http://dpdk.org/dev/patchwork/patch/11622/

### DIFF
--- a/dpdk/DPDK_17_02.vmware.patch
+++ b/dpdk/DPDK_17_02.vmware.patch
@@ -1,0 +1,22 @@
+--- lib/librte_eal/linuxapp/igb_uio/igb_uio.c
++++ lib/librte_eal/linuxapp/igb_uio/igb_uio.c
+@@ -388,14 +388,11 @@
+ 		}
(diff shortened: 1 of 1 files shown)


ifilippov:
Can you please explain what will this patch change (for what purpose) and why you don't apply it in Makefile?

sbilly:
It's to fix vmware e1000 input/output error. Please read : http://dpdk.org/dev/patchwork/patch/11622/

> In vmware esxi 6.0, add the emulated e1000 device to guest, then
> launch the VM, we will see the error meassage as below:
>
> EAL: Error reading from file descriptor 21: Input/output error
>
> INTX is badly emulated in Vmware, INTX toggle check didn't work
> with Vmware e1000 device. The patch is to fix the issue. And the
> patch is from http://dpdk.org/dev/patchwork/patch/945/, so Masaru
> OKI is the author. Qian just updated the patch for latest code.

ifilippov:
This patch was rejected by DPDK. I don't see any purpose to commit it here.

PR #28, opened 2017-06-23
Fix loss of starting packets in generated flow

mbulatova:
Delay is needed to let ports start. This is pktgen-like solution.


ifilippov:
Looks good to me, please commit.

PR #29, opened 2017-06-23
Add cmd line options to set port numbers in tests and examples

mbulatova:


ifilippov:
Gregory, do you think we need this?

gshimansky:
Can we have low-case command line switches in tests and examples? I like how it is done in test/performance/sendback.go.

mbulatova:
Added minor fix

PR #30, opened 2017-06-26
Fix pktgen download

mbulatova:
pktgen-v3.1.0 is now removed from dpdk.org. Move to next pktgen version


ifilippov:
1) Let's use 3.1.2 as latest 3.1 version.
2) Change our pktgen patch

ifilippov:
If building with these changes was tested, you can commit.

ifilippov:
Also you should change .gitignore files for new name of pktgen directory.

PR #31, opened 2017-06-27
Fix images generation for pktgen

gshimansky:
Pktgen image depends on yanff-base image, so yanff-base image should be
generated before dpdk, tests and examples images are built. This is done
by moving yanff-base to the upper level and setting dpdk dependency from
it. Tests and examples already depend on dpdk.


ifilippov:
Looks good to me

PR #33, opened 2017-06-29
Initialize payload length field for IPv6 protocols

gshimansky:
This is done for IPv4 but isn't done for IPv6 for some reason.


ifilippov:
I suppose that IPv6 payload field shouldn't contain size of IPv6 protocol.
Also what about InitEmptyEtherIPv6Packet function?

gshimansky:
Yes, for IPv6 this field doesn't include header length it seems.

ifilippov:
Looks good to me

PR #36, opened 2017-06-30
Some fix in packet pkg and add tests for InitEmpty func family

mbulatova:


ifilippov:
Looks good to me

ifilippov:
Please, commit.

PR #39, opened 2017-07-25
Added IPv4 ICMP protocol support

gshimansky:


packet/packet.go
```diff
@@ -169,6 +177,7 @@ type Packet struct {
 	IPv6     *IPv6Hdr       // Pointer to L3 header in mbuf (must be nil before parsing)
 	TCP      *TCPHdr        // Pointer to L4 header in mbuf (must be nil before parsing)
 	UDP      *UDPHdr        // Pointer to L4 header in mbuf (must be nil before parsing)
+	ICMP     *ICMPHdr       // Pointer to L4 header in mbuf (must be nil before parsing)
 	Data     unsafe.Pointer // Pointer to the packet payload data (invalid before parsing)
 	Unparsed uintptr        // Pointer whole packet data in mbuf (must be non-nil because it is the only public whole element)
```
ifilippov: Can you please add comment, that after adding one more protocol we will have problems with cache here.

packet/packet.go
```diff
@@ -154,6 +154,14 @@ func (hdr *UDPHdr) String() string {
...
 }
 
+type ICMPHdr struct {
+	Type       uint8  // ICMP message type
+	Code       uint8  // ICMP message code
+	Cksum      uint16 // ICMP checksum
+	Identifier uint16 // ICMP message identifier in some messages
+	SeqNum     uint16 // ICMP message sequence number in some messages
+}
+
```
ifilippov: Can you please add string method

ifilippov: Can you please add at the beginning of packet.go file comment, that we not support set of parsing functions for ICMP.

ifilippov:
Other looks good to me

packet/packet.go
ifilippov: If we have 9 protocols - we at least need to change mbufInit macros inside low.c

PR #46, opened 2017-07-28
Add checks of AppendMbuf return value

mbulatova:


ifilippov:
Not verbose but debug. Other looks good to me.

PR #50, opened 2017-08-10
First attempt to clean our packet structure

ifilippov:
As we handle only ethernet packets -> merge Unparsed and Ether fields of packet
It means:
1) packet structure is now inside one cache-line
2) ether field is parsed by default like "start of packet"
In API:
1) removed all ParseEther* functions -> parse ether by default
2) removed ParseL2 function -> parse ether by default
3) rename all InitEmptyEther* functions to InitEmpty*
4) removing of public Unparsed field. Users should use "packet.Start()"
function for pointer to first packet byte instead of Unparsed field
Really packet.Start function returns casted packet.Ether field and is inlined
5) ParseL3 returns length of L3 protocol
6) ParseL4 returns length of L3 + L4 protocol

### DIFF
--- a/low/low.c
+++ b/low/low.c
@@ -28,13 +28,16 @@
 	_mm_storeu_ps((float*)(buf + mbufStructSize + 16), zero128);
 #endif
 
-// This 56 offset is Unparsed field of packet struct. We can't non-set it here
-// because it can be different due to encapsulations and decapsulations.
+// This macros clears packet structure which is stored inside mbuf
+// mbufClear clears IPv4, IPv6, TCP and UDP protocols
+// 32 offset clears ICMP protocol
+// 40 offset is a data oddset and shouldn't be cleared
+// 48 offset is L2 offset and is always begining of packet
+// 56 offset is CMbuf offset and is initilized when mempool is created
 #define mbufInit(buf) \
 	mbufClear(buf) \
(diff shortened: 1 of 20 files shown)


low/low.c
```diff
@@ -27,13 +27,15 @@
 	_mm_storeu_ps((float*)(buf + mbufStructSize + 16), zero128);
 #endif
 
-// This 56 offset is Unparsed field of packet struct. We can't non-set it here
-// because it can be different due to encapsulations and decapsulations.
+// mbufClear clears IPv4, IPv6, TCP and UDP protocols
```
mbulatova: What exactly is cleared here?

low/low.c
```diff
@@ -279,8 +281,7 @@ struct rte_mempool * createMempool(uint32_t num_mbufs, uint32_t mbuf_cache_size)
 	temp = malloc(sizeof(struct rte_mbuf *) * num_mbufs);
 	allocateMbufs(mbuf_pool, temp, num_mbufs);
 	for (int i = 0; i < num_mbufs; i++) {
-		// TODO 64 const should be checked after all changes with packet struct
-		*(char**)((char*)(temp[i]) + mbufStructSize + 64) = (char*)(temp[i]);
+		*(char**)((char*)(temp[i]) + mbufStructSize + 56) = (char*)(temp[i]);
```
mbulatova: Can you pls explain this line? Maybe comment is needed here

PR #51, opened 2017-08-11
gofmt call - no changes

ifilippov:


TocarIP:
LGTM.
I'll always approve go fmt changes!

PR #54, opened 2017-08-17
Implementation of NAT example

gshimansky:


ifilippov:
Looks good to me. I'm not sure about whether we like to have file compiled only with future GO or not, but it looks good.

PR #55, opened 2017-08-25
To DPDK 17.08, pktgen 3.4.0

ifilippov:


ifilippov:
Performance and Stability were checked.

low/yanff_ring.h
```diff
@@ -17,7 +17,7 @@ yanff_ring_create(const char *name, unsigned count, int socket_id, unsigned flag
 	r = malloc(sizeof(struct yanff_ring));
 
 	r->DPDK_ring = rte_ring_create(name, count, socket_id, flags);
-	r->internal_DPDK_ring = r->DPDK_ring->ring;
-	r->offset = sizeof(r->DPDK_ring->ring[0]);
+	r->internal_DPDK_ring = &(r->DPDK_ring)[1];
```
mbulatova: Please clarify why such change. I see in dpdk 17.08 rte_ring has no void* ring field, but still the new value is not obvious

dpdk/DPDK_17_02.patch
```diff
@@ -1,18 +0,0 @@
...
- 
--	if (likely(rte_mbuf_refcnt_update(m, -1) == 0)) {
-+	// Here is YANFF change in DPDK. Original condition was
-+	// "rte_mbuf_refcnt_update(m, -1) == 0". However it reduced YANFF
-+	// performance because it wrote to the field refcnt in mbuf structure.
-+	// Changing this condition to reading is impossible if library uses
-+	// indirect mbufs due to data races. YANFF doesn't use
-+	// this functionality, so we can make this change. It is prohibited
-+	// to use this DPDK as normal DPDK after applying this patch!
-+	if (likely(rte_mbuf_refcnt_read(m) == 1)) {
```
mbulatova: Please explain briefly why with new dpdk the patch is not needed. Maybe commit description should say that.

low/yanff_ring.h
ifilippov: It is beginning of ring. Ring data is placed after rte_ring structure. So if we imagine that rte_ring is an array, so ring[1] is just after rte_ring structure itself. So it points just to beginning of ring data. I will add a comment.

dpdk/DPDK_17_02.patch
ifilippov: They change this function completely and it doesn't reduce our speed. Moreover now they also have rte_mbuf_refcnt_read as we are in our patch

PR #58, opened 2017-09-01
Misspells fix

aregm:
Fixed all misspells from goreport card.

### DIFF
--- a/low/low.go
+++ b/low/low.go
@@ -388,12 +388,12 @@ func yanff_ring_mc_do_dequeue(r *C.struct_yanff_ring, obj_table []uintptr, n uin
 	return n
 }
 
-// Heavilly based on DPDK mp_enqueue_burst
+// Heavily based on DPDK mp_enqueue_burst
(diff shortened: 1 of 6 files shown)


ifilippov:
Can you please merge all your changes into one commit because all of them all have one meaning

aregm:
You can squash them all to one commit during merge.

test/performance/latency/latency-part1.go
```diff
@@ -186,7 +186,6 @@ func latenciesLogger(ch <-chan time.Duration, stop <-chan string) {
 	for {
 		select {
 		case lat := <-ch:
-			sum += lat
```
gshimansky: Why did you remove this line?

ifilippov:
No. We need one commit with normal message like "Correction of misspells"

test/performance/latency/latency-part1.go
aregm: unused

ifilippov: What about sum := time.Duration(0) ?

aregm: Nice catch!

PR #59, opened 2017-09-01
A lot of NAT bugfixes

gshimansky:
Added check for TCP connection close.
Swap bytes of IP addresses.
Switched to two hash tables (improved performance and fixed correctness).
Improved cmd parameter parsing.
Improved debug diagnostics.
Fixed bug in pri2pub translation.


ifilippov:
IF OKEY -> THAN OKEY!

PR #70, opened 2017-09-14
Implement packet redesign

mbulatova:
Packet package:
Packet struct is now compact to fit into cache line.
- Packet struct now contains pointers to L3 and L4 headers,
instead of ptrs to all supported protocol headers
- Update Parse* function family which set L3 and L4 fields
- Add ParseAllKnownL3 and ParseAllKnownL4 which set L3 and L4 fields
and also return pointers to parsed headers
- Add Get* function family for cast parsed L3 and L4 to
certain header type
- Remove old protocol-specific Parse functions

Rules package:
- Now parsing of packet is done inside ACL functions

Other:
- Update examples and tests (internal, stability, performance)
according to new packet design

Authors:
Ilia Filippov <ilia.filippov@intel.com>,
Maria Bulatova <maria.bulatova@intel.com>,
Daria Kolistratova <daria.kolistratova@intel.com>


test/stability/test1/part1.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/intel-go/yanff/flow"
 	"github.com/intel-go/yanff/packet"
 
-	"github.com/intel-go/yanff/test/stability/test1/common"
+	"github.com/intel-go/yanff/test/stability/test1/test1Common"
```
ifilippov: What is it?

packet/packet.go
```diff
@@ -5,9 +5,9 @@
 // Package packet provides functionality for fast parsing and generating
 // of packets with known structure.
 // The following header types are supported:
-//	* L2 Ethernet
-//	* L3 IPv4 and IPv6
-//	* L4 TCP and UDP
+//      * L2 Ethernet
+//      * L3 IPv4 and IPv6
+//      * L4 TCP and UDP
```
ifilippov: Can you please add ICMP here

ifilippov:
I will check performance of this. We don't want to commit this before comming Labs.

low/low.c
```diff
@@ -29,15 +29,13 @@
 #endif
 
 // This macros clears packet structure which is stored inside mbuf
-// mbufClear clears IPv4, IPv6, TCP and UDP protocols
-// 32 offset clears ICMP protocol
-// 40 offset is a data oddset and shouldn't be cleared
-// 48 offset is L2 offset and is always begining of packet
-// 56 offset is CMbuf offset and is initilized when mempool is created
+// 0 offset is L3 protocol pointer
```
ifilippov: Can you please remove mbufClear Macros (22 string above).

test/stability/test1/part1.go
mbulatova: Just renaming of internal test file, so as not to confuse with github.com/intel-go/yanff/common package. It is not good if we have two packages with same names

ifilippov: Of course, however not in this commit... If it is not a problem, can you make a separate commit for it?

PR #74, opened 2017-09-26
Add support of dumping in pcap format to dump.go example

mbulatova:


ifilippov:
Maybe two options to one with non-bool parameter? Other great!

ifilippov:
Also what about SetWriter instead of SetReceiver? It won't be cloned and it is kind of OK?

mbulatova:
Yes SetWriter is unclonable like SetReceiver is. SetWriter is bound to fixed filename and is not supposed to clone. 
Here is just demonstration how pcap dumping can be used as clonable handler

examples/dump.go
```diff
@@ -35,16 +35,23 @@ func main() {
 	secondFlow := flow.SetPartitioner(firstFlow, 50000000, 1)
 
 	// Dump separated packet. By default function dumper() is used.
-	if *hexdumpOn {
-		flow.SetHandler(secondFlow, hexdumper, nil)
-	} else {
+	switch *dumptype {
+	case 1:
```
ifilippov: 1 for hex, not?

examples/dump.go
```diff
@@ -17,7 +17,7 @@ var (
 )
 
 func main() {
-	hexdumpOn := flag.Bool("hex", false, "enable dumping of packets in hex format")
+	dumptype := flag.Uint("dumptype", 0, "dumping format type (0 - dumper function, 1 - hex, 2 - pcap file)")
```
ifilippov: synchronize this with switch, please

ifilippov:
After correction of option handling you can merge - approved.

PR #76, opened 2017-09-29
Bugfix: segfault using L4 field of packet for icmp

dkolistr:
No Parse L4 was called


ifilippov:
Approved, please write informative commit message

PR #80, opened 2017-10-02
Added filling Proto field for IP6 protocol

dkolistr:
to init ipv6 header with data proto field should be set to "No Next
Header" it is 0x3B value. It was 0 which means "Hop-By-Hop Options Extension Header"  which must be supplied with some options, but I suppose if user calls InitEmptyIPv6Packet

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -26,10 +26,11 @@ const (
 
 // Supported L4 types
 const (
-	ICMPNumber = 0x01
-	IPNumber   = 0x04
-	TCPNumber  = 0x06
-	UDPNumber  = 0x11
+	ICMPNumber       = 0x01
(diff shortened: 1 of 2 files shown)


common/common.go
```diff
@@ -26,10 +26,11 @@ const (
...
 const (
-	ICMPNumber = 0x01
-	IPNumber   = 0x04
-	TCPNumber  = 0x06
-	UDPNumber  = 0x11
+	ICMPNumber       = 0x01
+	IPNumber         = 0x04
+	TCPNumber        = 0x06
+	UDPNumber        = 0x11
+	IPv6NoNextHeader = 0x59
```
mbulatova: You mean 0x3B ?

PR #81, opened 2017-10-02
Add perf tests for NAT example

mbulatova:


mbulatova:
Added tests with scheduler on/off, fixed/non-fixed ports

PR #82, opened 2017-10-02
Packet generation draft

dkolistr:


ifilippov:
Areg, can you please confirm that it is what you mentioned?

ifilippov:
Can you please change commit message to something more informative? Also I would like some instruction about this. Now it is hard to review.

PR #84, opened 2017-10-04
Fixed generate issue.

ifilippov:
Now fast generate functions try not to much exceed user provided speed
This is achieved by adding tiny pauses after packet processing inside generate


ifilippov:
Maria, can you please try this branch with your VMs and report results?

PR #85, opened 2017-10-05
Added checksum calculation to NAT. Changed checksum API

gshimansky:
Changed checksum API not to use getter functions inside


ifilippov:
Other tests are successful? Maria, can you please review this pull request.

Looks good to me.

mbulatova:
Tested branch, see incorrect HW checksums, reason is unknown.
While in test_cksums HW cksums are OK.

gshimansky:
Merged because checksums are required for correct TCP communication. We can fix HW checksums in a different commit.

PR #87, opened 2017-10-06
Add removing of user context after stopping a clone

ifilippov:
This functionality is useful for some freeing. Or is user wants to track number of clones. Now this functionality is full.

### DIFF
--- a/examples/clonable_pcap_dumper.go
+++ b/examples/clonable_pcap_dumper.go
@@ -59,6 +59,9 @@ func (pd pcapdumperParameters) Copy() interface{} {
 	return pdp
 }
 
+func (pd pcapdumperParameters) Delete() {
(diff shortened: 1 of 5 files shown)


ifilippov:
Maybe this functionality will be useful for NAT?

examples/clonable_pcap_dumper.go
```diff
@@ -59,6 +59,9 @@ func (pd pcapdumperParameters) Copy() interface{} {
 	return pdp
 }
 
+func (pd pcapdumperParameters) Remove() {
+}
```
mbulatova: File should be closed here

ifilippov: Can you please add this in separate commit

PR #89, opened 2017-10-06
Include cksum tests into stability.json

mbulatova:
Test is for HW checksums.
Correctness of SW checksums should be checked by external tool
like tcpdump


ifilippov:
Approved

PR #91, opened 2017-10-09
examples: fix issues pointed by "go lint"

Quasilyte:
Changes:
**1** - group imports into single `import (...)` clause
**2** - stdlib imports go first, then empty line, then other packages
**3** - `for true {}` should be written as `for {}`
**4** - made L3Rules global unexported (go lint)
**5** - `cloneNumber += 1` replaced with `cloneNumber++` (go lint)
**6** - omit type for `var` that is inferred from RHS (go lint)

**(1)** and **(3)** should be performed by `gofmt -s` (simplify code), but they are not, yet.
`goimports` is also sometimes refuses to re-arrange imports in a proper way, until you manually swap some lines.
The last 3 types of changes are caught by `go lint`.

**Note:** I have not looked into `examples/nat`, because of reasons.

### DIFF
--- a/examples/Firewall.go
+++ b/examples/Firewall.go
@@ -6,6 +6,7 @@ package main
 
 import (
 	"flag"
+
 	"github.com/intel-go/yanff/flow"
 	"github.com/intel-go/yanff/packet"
 	"github.com/intel-go/yanff/rules"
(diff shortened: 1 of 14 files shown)


ifilippov:
Approved, please merge.

Quasilyte:
@ifilippov, I am a little surprised if this is addressed to me, as I have no write access.

gshimansky:
No problem. I merged it for you.

PR #94, opened 2017-10-11
Fixed arp packets  and made small refactoring

dkolistr:
packet was filled wrong due to alignment, changed uint32 on [4]uint8, also
made common function instead of code duplication

### DIFF
--- a/packet/arp.go
+++ b/packet/arp.go
@@ -20,17 +20,17 @@ type ARPHdr struct {
 	PLen      uint8                      // Protocol address length, e.g. 4 for IPv4 address length
 	Operation uint16                     // Operation type, see ARP constants
 	SHA       [common.EtherAddrLen]uint8 // Sender hardware address (sender MAC address)
-	SPA       uint32                     // Sender protocol address (sender IPv4 address)
-	THA       [common.EtherAddrLen]uint8 // Target hardware address (target MAC address)
-	TPA       uint32                     // Target protocol address (target IPv4 address)
+	SPA       [common.IPv4AddrLen]uint8  // Sender protocol address (sender IPv4 address)
+	// array is used to avoid alignment (compiler alignes uint32 on 4 bytes)
+	THA [common.EtherAddrLen]uint8 // Target hardware address (target MAC address)
+	TPA [common.IPv4AddrLen]uint8  // Target protocol address (target IPv4 address)
+	// array is used to avoid alignment (compiler alignes uint32 on 4 bytes)
 }
 
 // ARP protocol operations
 const (
-	ARPRequest  = 1
-	ARPReply    = 2
-	RARPRequest = 3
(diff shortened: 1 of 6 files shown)


packet/arp.go
```diff
@@ -41,18 +43,18 @@ func (hdr *ARPHdr) String() string {
     PLen:  %d\n
     Operation: %d\n
     Sender MAC address: %02x:%02x:%02x:%02x:%02x:%02x\n
-    Sender IPv4 address: %d.%d.%d.%d\n
```
gshimansky: Changing IPv4 format from decimal to hex is wrong. Traditionally IPv4 addresses are printed in decimal.

packet/arp.go
```diff
@@ -67,29 +69,40 @@ func InitEmptyARPPacket(packet *Packet) bool {
...
-// InitARPRequestPacket initialize ARP request packet for IPv4
-// protocol request with broadcast (zero) for THA (Target HW
-// address). SHA and SPA specify sender MAC and IP addresses, TPA
-// specifies IP address for host which request is sent
-// for. Destionation MAC address is L2 Ethernet header is set to
-// FF:FF:FF:FF:FF:FF (broadcast) and source address is set to SHA.
-func InitARPRequestPacket(packet *Packet, SHA [common.EtherAddrLen]uint8, SPA, TPA uint32) bool {
+// InitARPCommonData allocates ARP packet, fills ether header and
+// arp hrd, pro, hln, pln with values for ether and IPv4
+func InitARPCommonData(packet *Packet) bool {
```
gshimansky: Let's make this function private to this file, i.e. from lower case letter. It is not needed from outside.

packet/arp.go
```diff
@@ -103,23 +116,14 @@ func InitARPRequestPacket(packet *Packet, SHA [common.EtherAddrLen]uint8, SPA, T
 // TPA specify target MAC and IP addresses. Destionation MAC address
 // is L2 Ethernet header is set to THA and source address is set to
 // SHA.
-func InitARPReplyPacket(packet *Packet, SHA, THA [common.EtherAddrLen]uint8, SPA, TPA uint32) bool {
-	var bufSize uint = common.EtherLen + common.ARPLen
-	if low.AppendMbuf(packet.CMbuf, bufSize) == false {
-		common.LogWarning(common.Debug, "InitGARPAnnouncementReplyPacket: Cannot append mbuf")
+func InitARPReplyPacket(packet *Packet, SHA, THA [common.EtherAddrLen]uint8, SPA, TPA [common.IPv4AddrLen]uint8) bool {
+	if !InitARPCommonData(packet) {
```
gshimansky: Please leave SPA and TPA parameter types as they were to uint32. Converting them to byte array should happen inside of service functions, not in public functions API.

PR #95, opened 2017-10-11
Added support for VLAN header

gshimansky:
Added functions to parse packets which may have VLAN header. More functions are necessary to generate such packets.

### DIFF
--- a/packet/packet.go
+++ b/packet/packet.go
@@ -72,12 +72,26 @@ type EtherHdr struct {
 }
 
 func (hdr *EtherHdr) String() string {
-	r0 := "L2 protocol: Ethernet\n"
+	r0 := fmt.Sprintf("L2 protocol: Ethernet\nEtherType: 0x%02x\n", hdr.EtherType)
 	s := hdr.SAddr
 	r1 := fmt.Sprintf("Ethernet Source: %02x:%02x:%02x:%02x:%02x:%02x\n", s[0], s[1], s[2], s[3], s[4], s[5])
 	d := hdr.DAddr
 	r2 := fmt.Sprintf("Ethernet Destination: %02x:%02x:%02x:%02x:%02x:%02x\n", d[0], d[1], d[2], d[3], d[4], d[5])
 	return r0 + r1 + r2
+}
+
+// VLANHdr 802.1Q VLAN header. We interpret it as an addition after
+// EtherHdr structure, so it contains actual frame EtherType after TCI
+// while TPID=0x8100 is present in EtherHdr.
+type VLANHdr struct {
+	TCI       uint16 // Tag control information. Contains PCP, DEI and VID bit-fields
+	EtherType uint16 // Real EtherType instead of VLANNumber in EtherHdr.EtherType
+}
+
+func (hdr *VLANHdr) String() string {
+	return fmt.Sprintf(`L2 VLAN:\n
+TCI: 0x%02x (priority: %d, drop %d, ID: %d)\n
+EtherType: 0x%02x\n`, hdr.TCI, byte(hdr.TCI>>13), (hdr.TCI>>12)&1, hdr.TCI&0xfff, hdr.EtherType)
 }
 
 // IPv4Hdr L3 header from DPDK: lib/librte_net/rte_ip.h
@@ -208,12 +222,43 @@ func (packet *Packet) Start() uintptr {
 	return uintptr(unsafe.Pointer(packet.Ether))
 }
 
-// ParseL3 set poinetr to start of L3 header
+// ParseL3 set pointer to start of L3 header
 func (packet *Packet) ParseL3() {
 	packet.L3 = unsafe.Pointer(packet.unparsed())
 }
 
-// GetIPv4 ensures if EtherType is IPv4 and cast L3 poinetr to IPv4Hdr type.
+// GetEtherType correctly returns EtherType from Ethernet header or
+// VLAN header.
+func (packet *Packet) GetEtherType() uint16 {
+	if packet.Ether.EtherType == SwapBytesUint16(VLANNumber) {
+		vptr := packet.unparsed()
+		vhdr := (*VLANHdr)(unsafe.Pointer(vptr))
+		return vhdr.EtherType
+	} else {
+		return packet.Ether.EtherType
+	}
+}
+
+// ParseL3 set pointer to start of L3 header taking possible presence
+// of VLAN header into account.
+func (packet *Packet) ParseL3CheckVLAN() {
+	ptr := packet.unparsed()
+
+	if packet.Ether.EtherType == SwapBytesUint16(VLANNumber) {
+		ptr += VLANLen
+	}
+	packet.L3 = unsafe.Pointer(ptr)
+}
+
+// GetVLAN returns VLAN header pointer if it is present in the packet.
+func (packet *Packet) GetVLAN() *VLANHdr {
+	if packet.Ether.EtherType == SwapBytesUint16(VLANNumber) {
+		return (*VLANHdr)(unsafe.Pointer(packet.unparsed()))
+	}
+	return nil
+}
+
+// GetIPv4 ensures if EtherType is IPv4 and cast L3 pointer to IPv4Hdr type.
 func (packet *Packet) GetIPv4() *IPv4Hdr {
 	if packet.Ether.EtherType == SwapBytesUint16(IPV4Number) {
 		return (*IPv4Hdr)(packet.L3)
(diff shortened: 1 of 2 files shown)


packet/packet.go
```diff
@@ -213,6 +226,53 @@ func (packet *Packet) ParseL3() {
...
+	} else {
+		packet.L3 = unsafe.Pointer(ptr)
+		return packet.Ether.EtherType, nil
+	}
+}
+
+// GetVLAN returns VLAN header pointer if it is present in the packet.
+func (packet *Packet) GetVLAN() *VLANHdr {
+	if packet.Ether.EtherType == SwapBytesUint16(VLANNumber) {
+		return (*VLANHdr)(unsafe.Pointer(uintptr(unsafe.Pointer(packet.Ether)) + EtherLen))
```
ifilippov: packet.Ether + EtherLen **-2** - no? VLAN tag not after etherHdr, but inside it, before etherType which is 2 byte long

packet/packet.go
```diff
@@ -213,6 +226,53 @@ func (packet *Packet) ParseL3() {
...
+	if packet.Ether.EtherType == SwapBytesUint16(VLANNumber) {
+		ptr += VLANLen
+	}
+	packet.L3 = unsafe.Pointer(ptr)
+}
+
+// ParseL3 set poinetr to start of L3 header taking possible VLAN
+// presence into account. It returns VLAN header if present and always
+// correct EtherType from VLAN header or Ethernet header.
+func (packet *Packet) ParseVLANAndL3() (uint16, *VLANHdr) {
```
ifilippov: I would like to name this function ParseL3ReturnVLAN. Also I suppose that this functions should return only vhdr. User may compare it to to nil after this function.

packet/packet.go
```diff
@@ -213,6 +226,53 @@ func (packet *Packet) ParseL3() {
...
+	ptr := uintptr(unsafe.Pointer(packet.Ether)) + EtherLen
+
+	if packet.Ether.EtherType == SwapBytesUint16(VLANNumber) {
+		vhdr := (*VLANHdr)(unsafe.Pointer(ptr))
```
ifilippov: ptr **-2** - no?

packet/packet.go
```diff
@@ -213,6 +226,53 @@ func (packet *Packet) ParseL3() {
 	packet.L3 = unsafe.Pointer(packet.unparsed())
 }
 
+// Correctly returns EtherType from Ethernet header or VLAN header.
```
ifilippov: Comment should begins with name of function

packet/packet.go
```diff
@@ -213,6 +226,53 @@ func (packet *Packet) ParseL3() {
...
+	if packet.Ether.EtherType == SwapBytesUint16(VLANNumber) {
+		vptr := uintptr(unsafe.Pointer(packet.Ether)) + EtherLen
+		vhdr := (*VLANHdr)(unsafe.Pointer(vptr))
+		return vhdr.EtherType
+	} else {
+		return packet.Ether.EtherType
+	}
+}
+
+// ParseL3 set poinetr to start of L3 header taking possible presence
```
ifilippov: Incorrect name of function and word "poinetr"

packet/packet.go
```diff
@@ -213,6 +226,53 @@ func (packet *Packet) ParseL3() {
...
+func (packet *Packet) ParseL3CheckVLAN() {
+	ptr := uintptr(unsafe.Pointer(packet.Ether)) + EtherLen
+
+	if packet.Ether.EtherType == SwapBytesUint16(VLANNumber) {
+		ptr += VLANLen
+	}
+	packet.L3 = unsafe.Pointer(ptr)
+}
+
+// ParseL3 set poinetr to start of L3 header taking possible VLAN
```
ifilippov: Incorrect name of function and word "poinetr"

packet/packet.go
```diff
@@ -221,14 +281,34 @@ func (packet *Packet) GetIPv4() *IPv4Hdr {
 	return nil
 }
 
-// GetIPv4 ensures if EtherType is IPv4 and cast L3 poinetr to IPv4Hdr type.
+// GetIPv4CheckVLAN ensures if EtherType is IPv4 and cast L3 poinetr
```
ifilippov: "poinetr"

packet/packet.go
```diff
@@ -213,6 +226,53 @@ func (packet *Packet) ParseL3() {
 	packet.L3 = unsafe.Pointer(packet.unparsed())
 }
 
+// Correctly returns EtherType from Ethernet header or VLAN header.
+func (packet *Packet) GetEtherType() uint16 {
+	if packet.Ether.EtherType == SwapBytesUint16(VLANNumber) {
+		vptr := uintptr(unsafe.Pointer(packet.Ether)) + EtherLen
```
ifilippov: Here and below: we have
"func (packet *Packet) unparsed() uintptr {
	return uintptr(unsafe.Pointer(packet.Ether)) + EtherLen
}"
for this common expression

ifilippov:
Please correct all comments. Also:

ParseL4ForIPv4, ParseL4ForIPv6 use packet.unparsed() and will both be incorrect. They should use "packet.L3" instead of "unparsed()" after this commit. User must parse L3 before L4 - no problems here. If user want - he would use ParseL3CheckVLAN.

Add function ParseAllKnownL3CheckVLAN which will use ParseL3CheckVLAN instead of ParseL3

Please send the result for one more review.

ifilippov:
hm, I understand why you don't use "-2" in ptr arithmetic. You suppose that first 0x8100 is part of etherHdr structure. Technically we can suppose so, however we need to point this clearly and externally. Because standard tells us that 802.1 Q has etherType as first byte and it is always 0x8100. Our structure representation where it is second byte will be very confusing. Please, add appropriate comments to etherHdr structure and VLANHdr structures

dkolistr:
Maybe it also makes sense to make functions for creation of vlan packets.

dkolistr:
Maybe also makes sense to add functions with creation of tagged packets. 
P.S. What if we have double tagging?

packet/packet.go
```diff
@@ -221,14 +281,34 @@ func (packet *Packet) GetIPv4() *IPv4Hdr {
...
+
+// GetARP ensures if EtherType is IPv4 and cast L3 poinetr to IPv4Hdr type.
 func (packet *Packet) GetARP() *ARPHdr {
 	if packet.Ether.EtherType == SwapBytesUint16(ARPNumber) {
 		return (*ARPHdr)(packet.L3)
 	}
 	return nil
 }
 
+// GetARPCheckVLAN ensures if EtherType is IPv4 and cast L3 poinetr to
```
mbulatova: "GetARPCheckVLAN ensures if EtherType is ARP and cast L3 pointer to ARPHdr type."

dkolistr: and castS

ifilippov:
Creation - other (next) commit. Double... hm

gshimansky:
I added a comment about our interpretation of VLANHdr. BTW, this is how DPDK interprets it, as an addition for EtherHdr, not a replacement of its part. I removed all unnecessary functions which I don't need, we can refine VLAN API later. So far I left only functionality that is needed for NAT implementation.

ifilippov:
Also:

ParseL4ForIPv4, ParseL4ForIPv6 use packet.unparsed() and will both be incorrect. They should use "packet.L3" instead of "unparsed()" after this commit. User must parse L3 before L4 - no problems here. If user want - he would use ParseL3CheckVLAN.

Add function ParseAllKnownL3CheckVLAN which will use ParseL3CheckVLAN instead of ParseL3

gshimansky:
I removed ParseL4ForIPv4, ParseL4ForIPv6. They are not present.

As for adding new functions, I prefer to refrain from adding many because it is not obvious yet which ones are going to be useful and which aren't. I am adding only those that are going to be used in practice.

dkolistr:
I need to create VLAN-tagged packets in test generator and it is very inconvenient now (as it will be for other users too). 
For example I use InitEmptyIPv4UDPPacket and similar functions for creation, but to make the same IPv4UDP packet with vlan tag I should init empty packet and do all by hand? 
I do not see any good solution instead of duplication a bunch of code making analogous  InitEmptyIPv4UDPPacketVLAN for each init function which adds 4 bytes to size, uses ParseL3CheckVLAN instead of ParseL3 and fills tag (which may come as parameter). But  duplication is always bad...does anyone have ideas how to make it in a better way?

ifilippov:
Daria, You can use our current functions and then EncapsulateHead(0, 4) and then some additional L2 parsing and settinf 0x8100 value. All other protocols will be at their places. So new functions will be {
currentFunction
Encapsulate
Parse
0x8100
}

Let's discuss this further by e-mails.

ifilippov:
Approved

PR #99, opened 2017-10-13
Added VLAN support to test generator

dkolistr:


ifilippov:
If you change only local test directory - then approved.

PR #100, opened 2017-10-15
Update README to require Go 1.9

nak3:
docs: Update README to require Go 1.9

In this project, `examples/nat/translation.go` uses sync.Map which has
been introduced on Go 1.9 - https://golang.org/doc/go1.9#sync-map. Due
to this, building fails with Go 1.8 env. This patch updates README to
require Go 1.9.

Signed-off-by: Kenjiro Nakayama <nakayamakenjiro@gmail.com>


aregm:
Good catch, thanks!

PR #101, opened 2017-10-16
Added comment and changed sum of pointer to Ether and ether len to

dkolistr:
packet.unparsed() in packet.AddVLANTag().


ifilippov:
Approved

PR #103, opened 2017-10-17
Added ARP support to NAT. Fixed checksum calculation in SW mode

gshimansky:
Successfully got a whole TCP session between a web browser and http server. First working NAT scenario!

### DIFF
--- a/packet/packet.go
+++ b/packet/packet.go
@@ -266,7 +266,7 @@ func (packet *Packet) GetIPv4() *IPv4Hdr {
 	return nil
 }
 
-// GetARP ensures if EtherType is IPv4 and cast L3 pointer to IPv4Hdr type.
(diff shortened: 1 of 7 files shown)


packet/packet.go
```diff
@@ -266,7 +266,7 @@ func (packet *Packet) GetIPv4() *IPv4Hdr {
 	return nil
 }
 
-// GetARP ensures if EtherType is IPv4 and cast L3 pointer to IPv4Hdr type.
+// GetIPv4 ensures if EtherType is ARP and cast L3 pointer to IPv4Hdr type.
```
ifilippov: Something is wrong:-)

ifilippov:
Approved. We merged first of two commits before, not?

gshimansky:
I decided to leave commits separate because they are not related. One is implementation of new feature (ARP support), another is bugfixes in checksums code.

PR #105, opened 2017-10-18
Implemented NAT reaction to ICMP echo request

gshimansky:
Now pings directed to NATs own IPs are returned back to sender to
allow network problems diagnostics.


ifilippov:
Approved

PR #106, opened 2017-10-19
Stability tests refactor and support run on VM

mbulatova:


ifilippov:
Why do you use delays? I supposed that we can delayed generates by scheduler.

mbulatova:
@ifilippov Start timeout is used to let generator stabilize speed. Packets sent and received
during timeout do not affect test result (because initial speed is high and a lot are dropped).
Description is corrected.

ifilippov:
Set schedSpeed to 1ms instead of making delays. This will reduce speed ~immediately. Can you try this?

PR #107, opened 2017-10-20
Encapsulate all translation data into one port-pair structure

gshimansky:
This allows parallel mutex free work of multiple port pairs on the
same NAT. Also code is more object oriented now.

Second commit fixes HW checksum calculation.


ifilippov:
Looks good to me. (No branch loops inside commit logic, yes?)

PR #110, opened 2017-10-30
Added support for VLAN enabled networks

gshimansky:


ifilippov:
Looks good to me, however: maybe divide for two commits? For packet.go and for NAT separately?
Commit name is bad, because we have already support VLANs, more details are required.

ifilippov:
And comments! First word in comment must be the same as function name!!!

examples/nat/main/Makefile
```diff
@@ -6,6 +6,6 @@ PATH_TO_MK = ../../../mk
 IMAGENAME = yanff-nat
 EXECUTABLES = nat
 
-nat: ../config.go ../translation.go ../portalloc.go
+nat: ../config.go ../translation.go ../portalloc.go ../cksum.go ../util.go
```
mbulatova: No file with name util.go added in this commit

gshimansky: Sorry, forgot to add it. It is fixed in pull request #111

PR #112, opened 2017-11-07
Added vagrant script to generate two VMs for YANFF tests

gshimansky:


ifilippov:
Approved (maybe vagrant inside test directory?)

PR #114, opened 2017-11-08
Update config2ports.json

scroot:
和结构体中的JSON序列化KEY不一致


gshimansky:
Hello. Why do you want to revert my change from "dst_mac" to "dst-mac"? I made this change to unify using hyphens instead of underscores like is is done in all other two-word parameters.

gshimansky:

PR #115, opened 2017-11-08
Update config.go

scroot:
同步修改结构体对应的JSON字段

### DIFF
--- a/examples/nat/config.go
+++ b/examples/nat/config.go
@@ -44,7 +44,7 @@ type portMapEntry struct {
 // Type describing a network port
 type ipv4Port struct {
 	Index         uint8      `json:"index"`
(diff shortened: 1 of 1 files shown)


gshimansky:
Hello. Why do you want to revert my change from "dst_mac" to "dst-mac"? I made this change to unify using hyphens instead of underscores like is is done in all other two-word parameters.

scroot:
1. 为了和DPDK保持习惯写法
2. 为了和之前的配置文件保持兼容
如果继续使用dst-mac的话，请把config.json同步修改一下

gshimansky:

Also, can you please write in English? I don't know Chinese, and I am not sure that google translates your messages adequately.

PR #135, opened 2017-11-21
Cosmetic corrections

gshimansky:


ifilippov:
dst-mac, really? (approved)

ifilippov:
To develop branch

examples/nat/main/config.json
```diff
@@ -8,7 +8,7 @@
             },
             "public-port": {
                 "index": 1,
-                "dst_mac": "3c:fd:fe:a4:d3:b1",
+                "dst-mac": "3c:fd:fe:a4:d3:b1",
```
mbulatova: Maybe remaining dst_mac -> dst-mac too?

gshimansky: dst_mac on private port should be deleted entirely since it is not used at all any more. Thanks for catching this.

gshimansky:
This change should not go into master yet

PR #139, opened 2017-11-27
PCAP utils refactor and unit tests

mbulatova:
Generalize pcap functions to accept io.Writer instead of *os.File


ifilippov:
Approved

PR #141, opened 2017-11-28
Added a new test on packet reordering

gshimansky:


mbulatova:
I suppose this two parts run on one machine, and on remote side some YANFF forwarding works? If so, maybe add file for remote side for completeness of test?

gshimansky:
The test may be used in any way with YANFF or Linux kernel between send and recv. I used this test to compare YANFF NAT with Linux kernel masquerading. So the middle program is arbitrary, whatever you think requires testing.

ifilippov:
Should we add run of this test to stability.json? Everything in stability directory is running via this file now. Approved.

gshimansky:
I don't have a scenario for this test to run, that is why I didn't make a framework config file for it.

PR #146, opened 2017-12-01
Fixed broken debug traces

gshimansky:
which broke because multiple packets were written simultaneously from
ingress and egress connection handlers or multiple threads.


ifilippov:
Approved

PR #148, opened 2017-12-04
Added error handling to some functions.

dkolistratova:
Major functions do not panic now, but return error.
Made special common error type, changed function calls in tests and examples, 
fixed name in arp test, magic number in packet utils test.
PLEASE read this really huge pull request, pay attention to files that you wrote to make sure I haven't broken anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,30 +24,42 @@ sacrificing performance.
 ### Feel the difference
 Simple ACL based firewall
 ```Go
+
+// CheckFatal is an error handling function
+func CheckFatal(err error) {
+	if err != nil {
+		fmt.Printf("checkfail: %+v\n", err)
+		os.Exit(1)
+	}
+}
+
 func main() {
 	// Initialize YANFF library to use 8 cores max.   
 	config := flow.Config{
 		CPUCoresNumber: 8,
 	}
-	flow.SystemInit(&config)
+	CheckFatal(flow.SystemInit(&config))
 
 	// Get filtering rules from access control file.
-	L3Rules = packet.GetL3ACLFromORIG("Firewall.conf")
+	L3Rules, err := packet.GetL3ACLFromORIG("Firewall.conf")
+	CheckFatal(err)
 
 	// Receive packets from zero port. Receive queue will be added automatically.
-	inputFlow := flow.SetReceiver(0)
+	inputFlow, err := flow.SetReceiver(uint8(0))
+	CheckFatal(err)
 
 	// Separate packet flow based on ACL.
-	rejectFlow := flow.SetSeparator(inputFlow, L3Separator, nil)
+	rejectFlow, err := flow.SetSeparator(inputFlow, L3Separator, nil)
+	CheckFatal(err)
 
 	// Drop rejected packets.
-	flow.SetStopper(rejectFlow)
+	CheckFatal(flow.SetStopper(rejectFlow))
 
 	// Send accepted packets to first port. Send queue will be added automatically.
-	flow.SetSender(inputFlow, 1)
+	CheckFatal(flow.SetSender(inputFlow, uint8(1)))
 
 	// Begin to process packets.
-	flow.SystemStart()
+	CheckFatal(flow.SystemStart())
 }
 
 // User defined function for separating packets
--- a/common/common.go
+++ b/common/common.go
@@ -7,12 +7,14 @@
 package common
 
 import (
-	"errors"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"runtime"
 	"strconv"
+
+	"github.com/pkg/errors"
 )
 
 // Length of addresses.
@@ -94,15 +96,153 @@ const (
 	TCPFlagCwr = 0x80
 )
 
+// ErrorCode type for codes of errors
+type ErrorCode int
+
+// constants with error codes
+const (
+	_ ErrorCode = iota
+	Fail
+	ParseCPUListErr
+	ReqTooManyPorts
+	BadArgument
+	UseNilFlowErr
+	UseClosedFlowErr
+	OpenedFlowAtTheEnd
+	PortHasNoQueues
+	NotAllQueuesUsed
+	FailToInitPort
+	ParseRuleJSONErr
+	FileErr
+	ParseRuleErr
+	IncorrectArgInRules
+	IncorrectRule
+	AllocMbufErr
+	PktMbufHeadRoomTooSmall
+	NotEnoughCores
+	CreatePortErr
+	MaxCPUExceedErr
+	PcapReadFail
+	PcapWriteFail
+	InvalidCPURangeErr
+)
+
+// NFError is error type returned by yanff functions
+type NFError struct {
+	Code     ErrorCode
+	Message  string
+	CauseErr error
+}
+
+type causer interface {
+	Cause() error
+}
+
+// Error method to implement error interface
+func (err NFError) Error() string {
+	return fmt.Sprintf("%s (%d)", err.Message, err.Code)
+}
+
+// GetNFErrorCode returns value of cCode field if err is
+// NFError or pointer to it and -1 otherwise.
+func GetNFErrorCode(err error) ErrorCode {
+	if nferr := GetNFError(err); nferr != nil {
+		return nferr.Code
+	}
+	return -1
+}
+
+func checkAndGetNFErrPointer(err error) *NFError {
+	if err != nil {
+		if nferr, ok := err.(NFError); ok {
+			return &nferr
+		} else if nferr, ok := err.(*NFError); ok {
+			return nferr
+		}
+	}
+	return nil
+}
+
+// GetNFError if error is NFerror or pointer to int
+// returns pointer to NFError, otherwise returns nil.
+func GetNFError(err error) (nferr *NFError) {
+	nferr = checkAndGetNFErrPointer(err)
+	if nferr == nil {
+		if cause, ok := err.(causer); ok {
+			nferr = checkAndGetNFErrPointer(cause.Cause())
+		}
+	}
+	return nferr
+}
+
+// Cause returns the underlying cause of error, if
+// possible. If not, returns err itself.
+func (err *NFError) Cause() error {
+	if err == nil {
+		return nil
+	}
+	if err.CauseErr != nil {
+		if cause, ok := err.CauseErr.(causer); ok {
+			return cause.Cause()
+		}
+		return err.CauseErr
+	}
+	return err
+}
+
+// Format makes formatted printing of errors,
+// the following verbs are supported:
+// %s, %v print the error. If the error has a
+// Cause it will be printed recursively
+// %+v - extended format. Each Frame of the error's
+// StackTrace will be printed in detail if possible.
+func (err *NFError) Format(s fmt.State, verb rune) {
+	switch verb {
+	case 'v':
+		if s.Flag('+') {
+			if cause := err.Cause(); cause != err && cause != nil {
+				fmt.Fprintf(s, "%+v\n", err.Cause())
+				io.WriteString(s, err.Message)
+				return
+			}
+		}
+		fallthrough
+	case 's', 'q':
+		io.WriteString(s, err.Error())
+	}
+}
+
+// WrapWithNFError returns an error annotating err with a stack trace
+// at the point WrapWithNFError is called, and the next our NFError.
+// If err is nil, Wrap returns nil.
+func WrapWithNFError(err error, message string, code ErrorCode) error {
+	err = &NFError{
+		CauseErr: err,
+		Message:  message,
+		Code:     code,
+	}
+	return errors.WithStack(err)
+}
+
 var currentLogType = No | Initialization | Debug
 
-// LogError internal, used in all packages
-func LogError(logType LogType, v ...interface{}) {
+// LogFatal internal, used in all packages
+func LogFatal(logType LogType, v ...interface{}) {
 	if logType&currentLogType != 0 {
 		t := fmt.Sprintln(v...)
 		log.Fatal("ERROR: ", t)
 	}
 	os.Exit(1)
+}
+
+// LogError internal, used in all packages
+func LogError(logType LogType, v ...interface{}) string {
+	if logType&currentLogType != 0 {
+		t := fmt.Sprintln(v...)
+		log.Print("ERROR: ", t)
+		return t
+	}
+	return ""
 }
 
 // LogWarning internal, used in all packages
@@ -167,12 +307,6 @@ func GetDefaultCPUs(cpuNumber uint) []uint {
 	return cpus
 }
 
-var (
-	// ErrMaxCPUExceed is error in case requested CPU exceeds maximum cores number on machine
-	ErrMaxCPUExceed    = errors.New("Requested cpu exceeds maximum cores number on machine")
-	ErrInvalidCPURange = errors.New("CPU range is invalid, min should not exceed max")
-)
-
 // ParseCPUs parses cpu list string into array of cpu numbers
 // and truncate the list according to given coresNumber (GOMAXPROCS)
 func ParseCPUs(s string, coresNumber uint) ([]uint, error) {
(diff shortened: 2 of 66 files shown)


scheduler/scheduler.go
```diff
@@ -322,7 +343,10 @@ func (scheduler *Scheduler) Schedule(schedTime uint) {
 }
 
 func (scheduler *Scheduler) startClone(ff *FlowFunction) bool {
-	index := scheduler.getCoreIndex(false)
+	index, err := scheduler.getCoreIndex(false)
+	if err != nil {
+		common.LogError(common.Debug, err)
+	}
 	if index != -1 {
```
mbulatova: I think this check is not necessary now. If err != nil we exit earlier on 348 line, otherwise index not -1

mbulatova: Also 'else' case below cannot ever execute

ifilippov:
Now you make the following:
1) Sprintf error message
2) pass it to Log function
3) pass it to return (sometimes)

I propose to do this shortly:
1) pass all components of error message to Log function. Let it return string of this message
2) use this string as you need.

This will significantly decrease amount of code.

I need to additionally check this commit after these (and above) corrections, please update it.

common/common.go
```diff
@@ -94,6 +94,44 @@ const (
 	TCPFlagCwr = 0x80
 )
 
+// ErrorCode type for codes of errors
+type ErrorCode int
+
+// constants with error codes
+const (
+	CodeOk ErrorCode = iota
+	CodeFail
```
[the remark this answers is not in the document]
dkolistr: yes, i just didn't want 0 to become error code

dkolistr:
pass all components of error message to Log function. Let it return string of this message
use this string as you need.

I dont want to return string, it

examples/tutorial/step11.go
```diff
@@ -16,19 +17,48 @@ const flowN = 3
...
-	flow.SetHandler(outputFlows[1], myHandler, nil)
+	firstFlow, err := flow.SetReceiver(0)
+	if err != nil {
+		log.Fatal(err)
+	}
+	outputFlows, err := flow.SetSplitter(firstFlow, mySplitter, flowN, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+	err = flow.SetStopper(outputFlows[0])
```
Quasilyte: It depends on the conventions, but I think that for error-only returns one can slightly simplify the code
and improve `err` variable locality by using init clause of `if`:
```go
// Before:
err = flow.SetHandler(outputFlows[1], myHandler, nil)
if err != nil {
	log.Fatal(err)
}
// After:
if err := flow.SetHandler(outputFlows[1], myHandler, nil); err != nil {
	log.Fatal(err)
}
```
It does not reduce real amount of code, but in my opinion it makes it easier to "scan through the code" with less clutter.

common/common.go
```diff
@@ -94,6 +94,42 @@ const (
...
+
+// NFError is error type returned by yanff functions
+type NFError struct {
+	Code    ErrorCode
+	Message string
+}
+
+// Error method to implement error interface
+func (err NFError) Error() string {
+	return fmt.Sprintf("failed with message '%s' and code '%d'", err.Message, err.Code)
```
Quasilyte: The rule of thumb for error messages for `Error` is that they should be easy to use in any logging scheme by end-user.
Proposing:
```go
return fmt.Sprintf("%s (%d)", err.Message, err.Code)
```
Let the user choose what to quote, or what quotes to use.
For example, your error could be used like this:
```go
if err != nil {
	log.Fatalf("failed at %v: yanff error='%v'", someContext, err)
}
```
Note the single quotes that wrap the error. With quotes inside Error return, output will be more ugly.

packet/pcap_utils.go
```diff
@@ -50,16 +50,22 @@ func WritePcapGlobalHdr(f io.Writer) {
...
-func (pkt *Packet) WritePcapOnePacket(f io.Writer) {
+func (pkt *Packet) WritePcapOnePacket(f io.Writer) error {
 	bytes := low.GetRawPacketBytesMbuf(pkt.CMbuf)
-	writePcapRecHdr(f, bytes)
-	writePacketBytes(f, bytes)
+	err := writePcapRecHdr(f, bytes)
+	if err != nil {
+		return err
+	}
+	err = writePacketBytes(f, bytes)
```
Quasilyte: Hmm.. maybe leaf error assignments should be turned into error returning statements?
```go
// Before:
err = writePacketBytes(f, bytes)
return err
// After:
return writePacketBytes(f, bytes)
```

scheduler/scheduler.go
dkolistr: I changed getCoreIndex, it can return -1, nil (if so, error is not fatal, else log error and exit)

common/common.go
```diff
@@ -105,6 +141,16 @@ func LogError(logType LogType, v ...interface{}) {
 	os.Exit(1)
 }
 
+// LogErrorNoExit internal, used in all packages
```
Quasilyte: This comment does not describe what function does.
That doc comment should say something about logging and string message return.

[Documenting go code](https://blog.golang.org/godoc-documenting-go-code).

Example:
```
LogErrorNoExit writes provided values to default logger if logType matches current log type.
Returns message that was printed.
```

Note: I know that common.go contains more such oddly documented functions that are out-of-scope for this review.

Quasilyte: If desired, "internal" package can be achieved since Go 1.5 with special import path:
```
An import of a path containing the element “internal” is disallowed if the importing code is outside the tree rooted at the parent of the “internal” directory.
```
This may require slight project layout re-organization, but it's more reliable way to make something "internal".

Quasilyte: What do you think about this:
1. rename `LogError` to `LogFatal`
2. rename `LogErrorNoExit` to `LogError`
?

ifilippov:
"Merge branch 'develop' into dkolistr/error_handling" - it will be looping here. You should use rebase, not merge. In merge case all develop changes will be from your branch and it will be hard to find actual changes.

ifilippov:
packet_test.go - group ifs to one line as above

low/low.go
```diff
@@ -503,11 +507,13 @@ func SetAffinity(coreID uint8) {
 }
 
 // AllocateMbufs allocates n mbufs.
-func AllocateMbufs(mb []uintptr, mempool *Mempool, n uint) {
+func AllocateMbufs(mb []uintptr, mempool *Mempool, n uint) error{
 	err := C.allocateMbufs((*C.struct_rte_mempool)(mempool), (**C.struct_rte_mbuf)(unsafe.Pointer(&mb[0])), C.unsigned(n))
 	if err != 0 {
-		common.LogError(common.Debug, "AllocateMbufs cannot allocate mbuf")
+		msg := common.LogErrorNoExit(common.Debug, "AllocateMbufs cannot allocate mbuf")
```
ifilippov: to one line?

flow/flow.go
```diff
@@ -1312,14 +1364,23 @@ func read(parameters interface{}, coreID uint8) {
 
 	// Read pcap global header once
 	var glHdr packet.PcapGlobHdr
-	packet.ReadPcapGlobalHdr(f, &glHdr)
+	err = packet.ReadPcapGlobalHdr(f, &glHdr)
```
ifilippov: One line?

ifilippov:
It will be impossible to write like Handle(Receive) now. Are we OK with this?

dkolistratova:
@ifilippov , I think yes, we have never used it in such a way.

common/common.go
dkolistratova: Yes, I think it should be refactored separately, this pull request is already too large.

dkolistratova:
@ifilippov, @gshimansky, what do u think about @Quasilyte proposal? 
rename LogError to LogFatal
rename LogErrorNoExit to LogError
?

ifilippov:
After function renaming approved, you can merge.

common/common.go
```diff
@@ -94,17 +94,63 @@ const (
...
+	return fmt.Sprintf("%s (%d)", err.Message, err.Code)
+}
+
 var currentLogType = No | Initialization | Debug
 
-// LogError internal, used in all packages
-func LogError(logType LogType, v ...interface{}) {
+// LogFatal internal, used in all packages
+func LogFatal(logType LogType, v ...interface{}) {
 	if logType&currentLogType != 0 {
```
aregm: There is an inconsistency: sometimes you use log.Fatal and sometimes LogFatal

examples/Forwarding.go
```diff
@@ -13,32 +15,50 @@ var l3Rules *packet.L3Rules
...
 		CPUList: "0-15",
 	}
-	flow.SystemInit(&config)
+	if err := flow.SystemInit(&config); err != nil {
+		log.Fatal(err)
+	}
 
 	// Get splitting rules from access control file.
-	l3Rules = packet.GetL3ACLFromORIG("Forwarding.conf")
+	l3Rules, err = packet.GetL3ACLFromORIG("Forwarding.conf")
```
aregm: UES

examples/Forwarding.go
```diff
@@ -13,32 +15,50 @@ var l3Rules *packet.L3Rules
...
+	if err != nil {
+		log.Fatal(err)
+	}
 
 	// Receive packets from zero port. Receive queue will be added automatically.
-	inputFlow := flow.SetReceiver(0)
+	inputFlow, err := flow.SetReceiver(uint8(0))
```
aregm: UES

examples/Forwarding.go
```diff
@@ -13,32 +15,50 @@ var l3Rules *packet.L3Rules
...
+	if err != nil {
+		log.Fatal(err)
+	}
 
 	// Split packet flow based on ACL.
 	flowsNumber := 5
-	outputFlows := flow.SetSplitter(inputFlow, l3Splitter, uint(flowsNumber), nil)
+	outputFlows, err := flow.SetSplitter(inputFlow, l3Splitter, uint(flowsNumber), nil)
```
aregm: UES

examples/antiddos/antiDDoS.go
```diff
@@ -81,15 +82,26 @@ func main() {
...
 	}
-	flow.SystemInit(&config)
+	if err := flow.SystemInit(&config); err != nil {
+		log.Fatal(err)
+	}
 
-	inputFlow := flow.SetReceiver(inPort)
-	flow.SetHandler(inputFlow, handle, nil)
-	flow.SetSender(inputFlow, outPort)
+	inputFlow, err := flow.SetReceiver(uint8(inPort))
```
aregm: UES

examples/antiddos/generatorForAntiDDoS.go
```diff
@@ -66,19 +67,37 @@ func main() {
...
 	exampleDoneEvent = sync.NewCond(&sync.Mutex{})
 
 	// Create first packet flow
-	outputFlow := flow.SetGenerator(generatePacket, speed, nil)
-	flow.SetSender(outputFlow, outPort)
-	inputFlow := flow.SetReceiver(inPort)
-	flow.SetHandler(inputFlow, checkInputFlow, nil)
-	flow.SetStopper(inputFlow)
+	outputFlow, err := flow.SetGenerator(generatePacket, speed, nil)
+	if err != nil {
```
aregm: UES

examples/copy.go
```diff
@@ -5,17 +5,32 @@
 package main
 
 import "github.com/intel-go/yanff/flow"
```
aregm: Order of the imports should be universal: system then local,  within parenthesis

README.md
```diff
@@ -29,25 +29,42 @@ func main() {
 	config := flow.Config{
 		CPUCoresNumber: 8,
 	}
-	flow.SystemInit(&config)
+	if err := flow.SystemInit(&config); err != nil {
```
aregm: I think you should wrap this into universal error handling function: CheckFatal(flow.SystemInit())

README.md
```diff
@@ -29,25 +29,42 @@ func main() {
...
+		log.Fatal(err)
+	}
 
 	// Get filtering rules from access control file.
-	L3Rules = packet.GetL3ACLFromORIG("Firewall.conf")
+	L3Rules, err := packet.GetL3ACLFromORIG("Firewall.conf")
+	if err != nil {
```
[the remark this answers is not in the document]
dkolistratova: L3Rules will be local if I do so, so either declare variables, or do as it is done.

common/common.go
```diff
@@ -94,17 +94,63 @@ const (
 	TCPFlagCwr = 0x80
 )
 
+// ErrorCode type for codes of errors
+type ErrorCode int
+
+// constants with error codes
+const (
+	CodeFail ErrorCode = iota + 1
```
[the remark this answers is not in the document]
dkolistratova: iota << 1 will be 0, 2, 4, ... i wanted 1, 2, 3, ... (starting from 1, cause don't want 0 to be one of error codes)

examples/Forwarding.go
dkolistratova: done in the form of
outputFlows, err := flow.SetSplitter(inputFlow, l3Splitter, uint(flowsNumber), nil)
CheckFatal(err)

examples/antiddos/antiDDoS.go
dkolistratova: the same

dkolistratova:
I made CheckFatal function in each file, because it is user-defined and can be different. It can check error code, act according to it, can use log.Fatal or anything else.

common/common.go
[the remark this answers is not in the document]
Quasilyte: It's common to have something like `no error` code.
```go
// constants with error codes
const (
	CodeOk ErrorCode = iota
	CodeFail
	// ... other codes
)
```
Or it's a bad idea in this case?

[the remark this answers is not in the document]
dkolistratova: @Quasilyte  It was exactly like this in previous version, but I got comment that this code is not needed for will never be used.

Quasilyte:
The https://github.com/intel-go/yanff/pull/171 can reduce error checking boilerplate a little.

Quasilyte:
It may take some effort to improve this, I am not sure it's ought to be in this very pull request.
It's a start that can be then be refactored/improved in certain way.

Before anything, I would like to see if each function really should return an error, or it's not that fatal to be reported as error. 
If all errors are significant, then we can make whole construction of objects somehow aggregated.
This can be done with construction functions ("constructors") that accept variadic number of "options" ([take inspiration here](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis)) or builder pattern. Either way, user will need to check initialization error exactly once.
Most setters can be turned into "options" (see article above).

For non-setters:
1. [text/scanner](https://golang.org/pkg/text/scanner/): allows user to set error handler as a struct field. If no handler is attached, it prints an error. We can choose "panic" instead of print.
The downside is that it is hard to handle some errors in this way (slightly better if each error has a good cause/error code).
2. [bufio/Scanner.Err](https://golang.org/pkg/bufio/#Scanner.Err): scanner holds an error and it can be queried with `Err()` later. If user does not care about errors, it does not call `Err()`.

dkolistratova:
@Quasilyte I made functions return error only in places where was Fatal logging (with exit), so they are significant.

Quasilyte:
@dkolistratova, for example, errors that show up due to invalid `interface{}` usage are not justified in my opinion.
https://github.com/intel-go/yanff/blob/f91e0e119df42831ae8d7a4033cb82e0e7da8b43/flow/flow.go#L661
`separateFunction` can be something else. For example, a named `Separator` interface.
How to do an interface that can support 2 functions? It may require interface-to-interface type assertion (used in stdlib to choose best Reader interface for the type) and interface embedding.
Or maybe it is possible to require single function with little re-think.
[go/importer](https://golang.org/pkg/go/importer/) does something like that with `Importer` and `ImporterFrom`.

With proper static type this function needs not to return an error.
But it's still better to turn all setters (or most of them) into builder or constructor.

dkolistratova:
Maybe, but it needs discussion with team and I suppose a separate pull request with redesign.

ifilippov:
I suppose that everything from pkg/errors is redundant and confusing for us. Also I do not welcome using separate packages.

However if we want it than I approve these changes because they seem to be implemented correctly.

PR #150, opened 2017-12-04
Fix some go vet warnings

mbulatova:


packet/vlan.go
```diff
@@ -56,15 +56,15 @@ func (packet *Packet) GetEtherType() uint16 {
 		vptr := packet.unparsed()
 		vhdr := (*VLANHdr)(unsafe.Pointer(vptr))
 		return SwapBytesUint16(vhdr.EtherType)
-	} else {
-		return SwapBytesUint16(packet.Ether.EtherType)
 	}
+	return SwapBytesUint16(packet.Ether.EtherType)
+
```
ifilippov: remove this line

ifilippov:
Can you please provide two directories: one with performance results before this and one with after. Not approved before this.

packet/vlan.go
gshimansky: This line is necessary because swapping EtherType is now made in GetEtherType()

ifilippov: Maybe. But empty line before the end of function cannot be necessary.

packet/packet.go
```diff
@@ -255,14 +257,12 @@ func (packet *Packet) GetIPv6NoCheck() *IPv6Hdr {
 
 // ParseL4ForIPv4 set L4 to start of L4 header, if L3 protocol is IPv4.
 func (packet *Packet) ParseL4ForIPv4() {
-	l3ptr := uintptr(unsafe.Pointer(packet.L3))
-	packet.L4 = unsafe.Pointer(l3ptr + uintptr((packet.GetIPv4NoCheck().VersionIhl&0x0f)<<2))
+	packet.L4 = unsafe.Pointer(uintptr(packet.unparsed()) + uintptr((packet.GetIPv4().VersionIhl&0x0f)<<2))
```
gshimansky: It is wrong to use GetIPv4 instead of GetIPv4NoCheck here. Function GetIPv4 can return nil and program will crash in this place. It happens for example if arrived packet is VLAN packet.

mbulatova:
Fixed remarks, provided perf results to @ifilippov, waiting for decision

ifilippov:
You can merge

PR #152, opened 2017-12-05
Changed debug trace to allow dumping traces into separate files

gshimansky:
Files are divided by interface type private/public. Old dump method of
mixing everything into one trace file is also still possible.


ifilippov:
Don't want to review "nat" changes. Always approved. Maybe you should develop in your own branch for some time...

PR #156, opened 2017-12-07
Fixes to NAT performance which finally make it faster than Linux kernel masquerading

gshimansky:


ifilippov:
Approved. I still suppose that you should develop in your own branch and commit only the final version as one commit instead of 70 incomplete commits.

PR #167, opened 2017-12-13
Improved performance system added.

ifilippov:
You should use
"make deploy"
and then
"make s1=machine1 s2=machine2 r=result perf_testing"
performance results will be in test/main/result_result file

### DIFF
--- a/test/main/perf_compare.go
+++ b/test/main/perf_compare.go
@@ -1,0 +1,97 @@
+// Copyright 2017 Intel Corporation.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+package main
+
+import (
+	"bufio"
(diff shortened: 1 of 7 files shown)


test/main/perf_compare.go
```diff
@@ -0,0 +1,97 @@
...
+			normal_cores := 6
+			if name[0] == "ipsec" {
+				normal_cores = 4
+			} else if name[0] == "nat_fixed_ports" || name[0] == "nat_nonfixed_ports" {
+				// Ask Maria!!!
+			}
+
+			j := i + 1
+			for strings.Contains(lines[j], "Average") == false {
+				j = j + 1
```
gshimansky: From here and to the end of function, why not use ++ operator for integer increment?

PR #168, opened 2017-12-13
Add example of gopacket use in handler

mbulatova:
Achieved speed on 64 byte packet is 1.2 Gb/s (scheduler off)
and 4.8 Gb/s (scheduler on)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,6 @@ examples/dump
 examples/clonable_pcap_dumper
 examples/kni
 examples/copy
+examples/gopacket_parser_example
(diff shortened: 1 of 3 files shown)


ifilippov:
Approved

mbulatova:
@aregm what do you think? should we add gopacket dependency to library (get_depends.sh) for this example to build it with all other examples? or leave it as standalone file like @ifilippov suggested?

mbulatova:
Example now can only be built separately by 'make gopacket_parser_example'

PR #170, opened 2017-12-14
Fix check for remote NUMA node warning

mbulatova:
Since binding the polling thread to any core is supported,
recv and send can run on NUMA node other than 0.
This commit switch on the warning for case of port on 0 numa node


ifilippov:
Approved

PR #172, opened 2017-12-19
Minor fix: set zero icmp checksum before calc

mbulatova:


gshimansky:
This is a side effect packet modification that I don't like. Checksum calculation function should not change the packet structure, instead the calling program should do it directly. Alternatively we can change the function contract to set icmp.Cksum to a new value.

mbulatova:
Ok, I think PR #173 is better solution. It is better to make these functions independent of Cksum field value, like other checksum functions are.

PR #174, opened 2017-12-19
Fix localTesting build

mbulatova:
Switched to new checksums API


ifilippov:
If it was checked - than approved.

PR #180, opened 2017-12-28
Update flow.go

scroot:
SetMerger 
checkFlow(InArray[i])


ifilippov:
Good catch! Thank you!

PR #187, opened 2018-01-18
Update .travis.yml and obsolete gopacket example

mbulatova:
Include example and localTesting dir into CI build


ifilippov:
Looks good to me

PR #188, opened 2018-01-18
flow: removed named results + minor stylistic changes

Quasilyte:
1. Instead of
```
x := new(T)
x.foo = y
```
I believe it is more readable to use keyed composite literal syntax:
```
x := &T{foo: y}
```

2. Named results should be avoided when they are not
strictly required (as in cases with defer+recover).
Also, I think that `out` name does not really add any documentation value, as `result` (output) parameters can't be anything else than outputs.
The `(T, error)` is very widely known, so `(t T, err error)` adds nothing valuable, too.

3. Renamed OUT to flow/out along the way, as all-upper is
not justified in this case (and violates Go conventions).
There are many more names like IN, but changing them in this
patch would make it harder for reviewer to inspect important changes.
Can be done later.

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -506,14 +506,13 @@ func SetWriter(IN *Flow, filename string) error {
 // file is read infinitely in circle.
 // Returns new opened flow with read packets.
 // Function can panic during execution.
-func SetReader(filename string, repcount int32) (OUT *Flow) {
+func SetReader(filename string, repcount int32) *Flow {
 	ring := low.CreateRing(generateRingName(), burstSize*sizeMultiplier)
 	read := makeReader(filename, ring, repcount)
 	schedState.UnClonable = append(schedState.UnClonable, read)
-	OUT = new(Flow)
-	OUT.current = ring
+	flow := &Flow{current: ring}
 	openFlowsNumber++
-	return OUT
+	return flow
 }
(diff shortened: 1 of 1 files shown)


Quasilyte:
Also, types like `[N](*T)` can be simplified to `[N]*T`.

ifilippov:
1) "flow" is a bad name, because there are lots of flows there. Let it be outFlow and outFlowArray if you don't like OUT. Also you should change IN to inFlow in this naming notation.
2) I'm strictly against "return &Flow{current: ringSecond}" notation. Current version seems more readable.
3) [N](*T) is more readable than [N]*T because it requires less time to understand.
4) Let's have these changes (except points 2 and 3) after Maria's changes because there will be a lot of rebasing and Maria's changes are more important.

Quasilyte:
Thank you for quick response and review.

> Also you should change IN to inFlow in this naming notation.

I tried to concentrate only on out-params, as renaming of input params would also require doc-comments update.

> [N] (*T) is more readable than [N]*T because it requires less time to understand.

I think there was some rule about that, but I can't find it right now, so my argument currently has no objective strength.
Made an issue here though:
https://github.com/dominikh/go-tools/issues/254
Go core team could response there. I could be wrong all this time..

PR #189, opened 2018-01-18
Enabling GTPv1-U

ifilippov:
No Testing, even dumping. Just prototype.

### DIFF
--- a/packet/gtp.go
+++ b/packet/gtp.go
@@ -1,0 +1,138 @@
+// Copyright 2017 Intel Corporation.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// This file highly assumes that all packets in each particular situation
+// will have the same structure. Parsing, Encapsulating, Decapsulating
+// functions are base variants without any checkings which can be
+// painlessly changed for exact situation.
+
+package packet
+
+import (
+	"fmt"
+	"unsafe"
+
+	. "github.com/intel-go/yanff/common"
+)
+
+// GTPv1-U and GTPv1-C
+type GTPHdr struct {
+	HeaderType          uint8 // version, protocol type, extension header/sequence number/N-PDU flags
+	MessageType         uint8
+	MessageLength       uint16
+	TEID                uint32
+	SequenceNumber      uint16 // this is valid only with sequence number flag
+	NPDUNumber          uint8  // this is valid only with N-PDU flag
+	NextExtensionHeader uint8  // this is valid only with exatension header flag
+}
+
+// TODO current support of GTP-U: only two extension headers
+// NextExtensionHeader values
+const (
+	NoExtensionHeaders             = 0x00
+	UDPPortExtensionHeader         = 0x40
+	PDCP_PDU_NumberExtensionHeader = 0xc0
+)
+
+type UDPPort struct {
+	Lenght              uint8 // in 4 octets, here always 0x01
+	UDPPortNumber       uint16
+	NextExtensionHeader uint8
+}
+
+type PDCP_PDU_Number struct {
+	Length              uint8 // in 4 octets, here always 0x01
+	PDCP_PDU_Number_1   uint8
+	PDCP_PDU_Number_2   uint8
+	NextExtensionHeader uint8
+}
+
+// TODO for GTP-U only
+// MessageType values
+const (
+	// Signalling messages: path management
+	EchoRequest                           = 1
+	EchoResponse                          = 2
+	SupportedExtensionHeadersNotification = 31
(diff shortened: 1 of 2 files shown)


packet/gtp.go
```diff
@@ -0,0 +1,138 @@
...
+// functions are base variants without any checkings which can be
+// painlessly changed for exact situation.
+
+package packet
+
+import (
+	"fmt"
+	"unsafe"
+
+	. "github.com/intel-go/yanff/common"
```
aregm: Better to do alias and we need to solve the "common.go" problem

packet/gtp.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+// GTPv1-U and GTPv1-C
+type GTPHdr struct {
+	HeaderType          uint8 // version, protocol type, extension header/sequence number/N-PDU flags
+	MessageType         uint8
+	MessageLength       uint16
+	TEID                uint32
+	SequenceNumber      uint16 // this is valid only with sequence number flag
```
aregm: remove "this is"

packet/gtp.go
Quasilyte: Usage of dot imports outside of testing code is discouraged:
See https://github.com/golang/go/wiki/CodeReviewComments#import-dot.

Quasilyte: Many go static code analyzers can't handle dot-imports.
Also, `go lint` will say: `should not use dot imports`.

ifilippov:
Quasilyte: We use dot import from common.go across whole framework due to significant code reducing. Maybe it can be refactored somehow but not in this commit.

Quasilyte:
You mentioned that this is "prototyping", so it was quite wrong to comment on this.
I done this before reading the actual description, sorry.

packet/gtp.go
```diff
@@ -0,0 +1,138 @@
...
+	EchoResponse                          = 2
+	SupportedExtensionHeadersNotification = 31
+	// Signalling messages: tunnel management
+	ErrorIndication = 26
+	EndMarker       = 254
+	// G-PDU message = encapsulated user data
+	G_PDU = 255
+)
+
+// TODO Add "Information Element@ constants
```
aregm: What does comment want to say?

packet/gtp.go
```diff
@@ -0,0 +1,138 @@
...
+	Length              uint8 // in 4 octets, here always 0x01
+	PDCP_PDU_Number_1   uint8
+	PDCP_PDU_Number_2   uint8
+	NextExtensionHeader uint8
+}
+
+// TODO for GTP-U only
+// MessageType values
+const (
+	// Signalling messages: path management
```
aregm: Alignment

packet/gtp.go
```diff
@@ -0,0 +1,138 @@
...
+		min += fmt.Sprintf(`, Sequence number %d`, SwapBytesUint16(hdr.SequenceNumber))
+	}
+	if hdr.HeaderType&0x01 != 0 {
+		min += fmt.Sprintf(`, N-PDU number %d`, hdr.NPDUNumber)
+	}
+	if hdr.HeaderType&0x04 != 0 {
+		min += fmt.Sprintf(`, N-PDU number %d`, hdr.NextExtensionHeader)
+		// TODO add dumping of all extension headers
+	}
+	return min
```
aregm: Why name 'min'?

packet/gtp.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// EncapsulateIPv4GTP assumes that user wants to build ether->IPv4->UDP->GTP->payload data structure
+// with standart IPv4 header size. It is also assumed that payload type is IPv4, so no etherType changes are needed
+func (packet *Packet) EncapsulateIPv4GTP(TEID uint32) bool {
+	if !packet.EncapsulateHead(EtherLen, IPv4MinLen+UDPLen+GTPMinLen) {
+		return false
+	}
+	gtp := (*GTPHdr)(unsafe.Pointer(uintptr(packet.unparsed()) + IPv4MinLen + UDPLen))
+	gtp.HeaderType = 0x30   // 001 - GTPv1, 1 - not GTP', 0 - reserved, 000 - no optional fields
```
aregm: Better to define constant

packet/gtp.go
ifilippov: It is comment for underline variable

packet/gtp.go
ifilippov: That we need to add "Information Element" constants which are defined in protocol RFCs

packet/gtp.go
ifilippov: Because it is minimal protocol length

PR #191, opened 2018-01-22
Clearing of tests

ifilippov:
GetPacket everywhere
package packet everywhere
removal of unnecessary functionality

### DIFF
--- a/packet/checksum_test.go
+++ b/packet/checksum_test.go
@@ -2,18 +2,17 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package packet_test
+package packet
 
 import (
 	"encoding/binary"
 	"github.com/intel-go/yanff/common"
-	"github.com/intel-go/yanff/packet"
 	"net"
 	"testing"
 )
 
 const N uint = 20
-const payloadSize = N * 8
+const payloadSizeLocal = N * 8
 
 type Payload struct {
 	array [N]uint64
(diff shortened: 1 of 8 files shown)


packet/utils_for_test.go
```diff
@@ -11,38 +11,25 @@ import (
...
-	if testMempool != nil {
-		GetMempoolForTest()
-	}
-}
-
-// GetMempoolForTest initialize DPDK and testMempool.
-// Calling this function guarantee initialization is done once during all tests in package.
-func GetMempoolForTest() *low.Mempool {
-	if testMempool == nil {
+func TInitDPDK() {
```
dkolistratova: Comment to exported function is required by golint. Maybe better Test instead of T?

packet/checksum_test.go
```diff
@@ -39,146 +38,134 @@ CalculatePseudoHdrIPv6UDPCksum
...
 		t.Errorf("Incorrect result:\ngot: %x, \nwant: %x\n\n", got, want)
 	}
 }
 
 func TestCalculateIPv6UDPChecksum(t *testing.T) {
-	pkt := packet.GetPacket()
-	packet.InitEmptyIPv6UDPPacket(pkt, payloadSize)
-	initIPv6Addrs(pkt)
+	pkt := GetPacket()
+	InitEmptyIPv6UDPPacket(pkt, payloadSize2)
```
dkolistratova: awful name! maybe better add local or smth making sence

packet/checksum_test.go
```diff
@@ -39,146 +38,134 @@ CalculatePseudoHdrIPv6UDPCksum
...
 	}
 }
 
 func TestCalculateIPv6ICMPChecksum(t *testing.T) {
-	pkt := packet.GetPacket()
-	packet.InitEmptyIPv6ICMPPacket(pkt, payloadSize)
-	initIPv6Addrs(pkt)
+	pkt := GetPacket()
+	InitEmptyIPv6ICMPPacket(pkt, payloadSize2)
+	initIPv6Addrs2(pkt)
```
dkolistratova: again this "2"

packet/checksum_test.go
```diff
@@ -39,146 +38,134 @@ CalculatePseudoHdrIPv6UDPCksum
...
 	}
 }
 
-func initEtherAddrs(pkt *packet.Packet) {
-	pkt.Ether.SAddr = [common.EtherAddrLen]byte{0x11, 0x22, 0x33, 0x44, 0x55, 0x66}
-	pkt.Ether.DAddr = [common.EtherAddrLen]byte{0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f}
-}
-
-func initIPv4Addrs(pkt *packet.Packet) {
+func initIPv4Addrs2(pkt *Packet) {
```
dkolistratova: such amount of "2" functions and variables is abnormal
maybe  common function with parameters is needed?

packet/checksum_test.go
```diff
@@ -2,18 +2,17 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package packet_test
+package packet
```
mbulatova: It is not very critical, I am just not sure we need make all tests internal.
In this file we have tests for only API functions, it should be ok to leave it packet_test as assumed for API tests.

packet/checksum_test.go
mbulatova: > maybe common function with parameters is needed?

exported, placed in utils_for_test.go.

packet/packet_test.go
```diff
@@ -2,27 +2,22 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package packet_test
+package packet
```
mbulatova: The same as above

packet/checksum_test.go
ifilippov: We don't want to have multiple packages

packet/utils_for_test.go
ifilippov: Test__ is a function for testing? it should have different signature

packet/checksum_test.go
ifilippov: I will add "l" instead of 2.

packet/checksum_test.go
ifilippov: That functions is needed only in acl_test. I don't know why they are in utils really, however It is strange idea to make functions that only initialize values with parameters. I will place "l" instead of "2"

packet/utils_for_test.go
ifilippov: This shouldn't be exported. As well as GetPacket. Will correct this.

ifilippov:
Updated, please comment.

packet/utils_for_test.go
mbulatova: It is not obvious, but it is ok to export in this file
Directive // +build !test 
makes exported functions present only if build for testing

dkolistratova:
Please change L to Local (L is not obvious) and then commit, we'll think of a better solution and send a separate pull request.

PR #193, opened 2018-01-24
Updated tutorial files, added presentation

gshimansky:


ifilippov:
The main point in this pull request is usage of atomics in some examples connected with updating ACL rules dynamically. Performance was tested - no performance degradation, so approved.

PR #198, opened 2018-01-30
Update Makefile

aregm:
Add non-parallel testing option to overcome hugepages error on Travis CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,8 @@ perf_testing:
 	$(MAKE) -C test perf_testing
 
 .PHONY: testing $(TESTING_TARGETS)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -19,7 +19,9 @@ perf_testing:
 	$(MAKE) -C test perf_testing
 
 .PHONY: testing $(TESTING_TARGETS)
-testing: $(TESTING_TARGETS)
+.NOTPARALLEL: testing $(TESTING_TARGETS)
+testing $(TESTING_TARGETS)
```
gshimansky: Here should be "testing: $(TESTING_TARGETS)" like it was before the change.

PR #199, opened 2018-01-31
Refactor of CPU list handling

mbulatova:
SetAffinity now returns error.
ParseCPU now does only parsing. Add DropInvalidCPUs.
Common test now does not not require many cores.

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -12,7 +12,6 @@ import (
 	"log"
 	"math"
 	"os"
-	"runtime"
 	"strconv"
 
 	"github.com/pkg/errors"
@@ -132,6 +131,7 @@ const (
 	PcapReadFail
 	PcapWriteFail
 	InvalidCPURangeErr
+	SetAffinityErr
 )
 
 // NFError is error type returned by yanff functions
@@ -305,7 +305,7 @@ func GetDPDKLogLevel() string {
 	}
 }
 
-// GetDefaultCPUs returns default core list {0, 1, ..., GOMAXPROCS-1}
+// GetDefaultCPUs returns default core list {0, 1, ..., NumCPU}
 func GetDefaultCPUs(cpuNumber uint) []uint {
 	cpus := make([]uint, cpuNumber, cpuNumber)
 	for i := uint(0); i < cpuNumber; i++ {
@@ -314,9 +314,17 @@ func GetDefaultCPUs(cpuNumber uint) []uint {
 	return cpus
 }
 
-// ParseCPUs parses cpu list string into array of cpu numbers
-// and truncate the list according to given coresNumber (GOMAXPROCS)
-func ParseCPUs(s string, coresNumber uint) ([]uint, error) {
+// HandleCPUs parses cpu list string into array of valid core numbers.
+// Removes duplicates
+func HandleCPUList(s string, maxcpu uint) ([]uint, error) {
+	nums, err := parseCPUs(s)
+	nums = removeDuplicates(nums)
+	nums = dropInvalidCPUs(nums, maxcpu)
+	return nums, err
+}
+
+// parseCPUs parses cpu list string into array of cpu numbers
+func parseCPUs(s string) ([]uint, error) {
 	var startRange, k int
 	nums := make([]uint, 0, 256)
 	if s == "" {
@@ -355,19 +363,6 @@ func ParseCPUs(s string, coresNumber uint) ([]uint, error) {
 			j = i + 1
 		}
 	}
-
-	nums = removeDuplicates(nums)
-
-	numCPU := uint(runtime.NumCPU())
-	for _, cpu := range nums {
-		if cpu >= numCPU {
-			return []uint{}, WrapWithNFError(nil, "requested cpu exceeds maximum cores number on machine", MaxCPUExceedErr)
-		}
-	}
-	if len(nums) > int(coresNumber) {
-		return nums[:coresNumber], nil
-	}
-
 	return nums, nil
 }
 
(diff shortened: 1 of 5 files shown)


common/common_test.go
```diff
@@ -18,41 +18,49 @@ var invalidCpuRangeErr = common.WrapWithNFError(nil, "CPU range is invalid, min
...
+	{"10-14,11-12", []uint{10, 11, 12, 13, 14}, nil},
+	{"1-3,6-", []uint{1, 2, 3}, &strconv.NumError{"Atoi", "", strconv.ErrSyntax}},
+	{"-1", []uint{}, &strconv.NumError{"Atoi", "", strconv.ErrSyntax}},
+	{"10-6", []uint{}, common.GetNFError(invalidCpuRangeErr)},
+	{"1-3,10-6", []uint{1, 2, 3}, common.GetNFError(invalidCpuRangeErr)},
 }
 
-// TestParseCPUs require minumum 22 cores to pass without error.
-// If machine has less cores, test fail with ErrMaxCPUExceed error
 func TestParseCPUs(t *testing.T) {
```
dkolistratova: I think golint may swear at exported function with no comment.

common/common_test.go
```diff
@@ -18,41 +18,49 @@ var invalidCpuRangeErr = common.WrapWithNFError(nil, "CPU range is invalid, min
...
+var dropInvalidCPUsTests = []struct {
+	cpus     []uint
+	maxcpu   uint
+	expected []uint // expected valid cpu list
+}{
+	{[]uint{}, 20, []uint{}},
+	{[]uint{1, 2, 100, 1, 2, 100}, 20, []uint{1, 2, 1, 2}},
+}
+
+// TestParseCPUs require minumum 22 cores to pass without error.
```
dkolistratova: TestDropInvalidCPUs

flow/flow.go
```diff
@@ -358,9 +358,10 @@ func SystemInit(args *Config) error {
 	var cpus []uint
 	var err error
 	if args.CPUList != "" {
-		if cpus, err = common.ParseCPUs(args.CPUList, CPUCoresNumber); err != nil {
+		if cpus, err = common.ParseCPUs(args.CPUList); err != nil {
 			return err
 		}
+		cpus = common.DropInvalidCPUs(cpus, CPUCoresNumber)
```
dkolistratova: Please ask Ilya or Gregory about CPUCoresNumber as max cpu number?

scheduler/scheduler.go
```diff
@@ -156,15 +156,16 @@ func NewScheduler(cpus []uint, schedulerOff bool, schedulerOffRemove bool,
...
 	err = common.WrapWithNFError(nil, "Requested number of cores isn't enough. System needs at least one core per each Set function (except Merger and Stopper) plus one additional core.", common.NotEnoughCores)
 	index := scheduler.getCoreIndex()
 	if index == -1 {
 		return err
 	}
 	core := scheduler.cores[index].id
 	common.LogDebug(common.Initialization, "Start SCHEDULER at", core, "core")
-	low.SetAffinity(uint8(core))
+	if err := low.SetAffinity(uint8(core)); err != nil {
+		return err
```
dkolistratova: I think we may wrap it with common.WrapWithNFError

scheduler/scheduler.go
```diff
@@ -176,7 +177,10 @@ func (scheduler *Scheduler) SystemStart() (err error) {
 		common.LogDebug(common.Initialization, "Start STOP at scheduler", core, "core")
 	}
 	go func() {
-		low.SetAffinity(uint8(core))
+		err := low.SetAffinity(uint8(core))
+		if err != nil {
+			common.LogFatal(common.Initialization, "Failed to set affinity to", core, "core: ", err)
```
dkolistratova: maybe return common.WrapWithNFError instead of fatal logging?

dkolistratova: here and below

mbulatova: I place LogFatal only in goroutines, because we do not return values from goroutines

mbulatova: Here are several goroutines, each can give this error, maybe we should think of writing its errors to global channel or err variable?

common/common_test.go
ifilippov: No, this is just test function in test file

scheduler/scheduler.go
ifilippov: LogFatal only in goroutines -> it isn't right logic. Right logic is only via execution. Here is not execution - it is initialization of stop goroutine. So you can return error (one more time our errors are awful) with comment about "for stopping thread"

low/low.go
```diff
@@ -507,14 +507,18 @@ func CreateMempool() *Mempool {
...
 	runtime.LockOSThread()
 
 	var cpuset C.cpu_set_t
 	C.initCPUSet(C.uint8_t(coreID), &cpuset)
-	syscall.RawSyscall(syscall.SYS_SCHED_SETAFFINITY, uintptr(0), unsafe.Sizeof(cpuset), uintptr(unsafe.Pointer(&cpuset)))
+	_, _, errno := syscall.RawSyscall(syscall.SYS_SCHED_SETAFFINITY, uintptr(0), unsafe.Sizeof(cpuset), uintptr(unsafe.Pointer(&cpuset)))
+	if errno != 0 {
+		return common.WrapWithNFError(nil, errno.Error(), common.SetAffinityErr)
+	}
+	return nil
```
ifilippov: Was this tested?

common/common.go
```diff
@@ -382,3 +371,16 @@ func removeDuplicates(array []uint) []uint {
...
+func DropInvalidCPUs(nums []uint, maxcpu uint) []uint {
+	i := 0
+	for _, x := range nums {
+		if x < maxcpu {
+			nums[i] = x
+			i++
+		}
+	}
+	return nums[:i]
+}
```
ifilippov: Really? Without any notice to user?

ifilippov:
Small refactor? Really? Ideal name

low/low.go
mbulatova: Yes, I made it to fail and it fails as expected

flow/flow.go
ifilippov: NumCPU - OK, however I suppose that all user input should be handled in one function. Parsing, duplication, trunking, any other points.

scheduler/scheduler.go
mbulatova: So we need to return value from running goroutine. It seems correct way is to use channel https://www.atatus.com/blog/goroutines-error-handling/

ifilippov:
Approved. Name it "Refactor of CPU list handling"

PR #200, opened 2018-02-01
Switched private and public network ports for convenince

gshimansky:


ifilippov:
Approved, however I suppose that NAT commits should use separate branch instead of merging them directly into develop.

PR #204, opened 2018-02-06
Rewritten tests to make them run with go test

dkolistratova:


gshimansky:
Since you are changing all tests anyway, let's get rid of CPU cores requirement. In my branch gregory/testing I removed all `CPUList` assignments because we don't have 16 CPU cores in VMs and none of these tests requires so many cores. Let's keep `CPUList` unassigned to let tests use defaults.

test/stability/test_merge/main-test-merge.go
```diff
@@ -75,55 +83,96 @@ func CheckFatal(err error) {
...
 	flag.UintVar(&outport1, "outport1", 0, "port for 1st sender")
 	flag.UintVar(&outport2, "outport2", 1, "port for 2nd sender")
-	flag.UintVar(&inport, "inport", 0, "port for receiver")
+	flag.UintVar(&inport1, "inport1", 0, "port for 1st receiver")
+	flag.UintVar(&inport2, "inport2", 1, "port for 2nd receiver")
 	flag.Uint64Var(&totalPackets, "number", totalPackets, "total number of packets to receive by test")
 	flag.DurationVar(&T, "timeout", T, "test start delay, needed to stabilize speed. Packets sent during timeout do not affect test result")
 	configFile := flag.String("config", "", "Specify json config file name (mandatory for VM)")
-	target := flag.String("target", "", "Target host name from config file (mandatory for VM)")
+	target := flag.String("target", "hostnme2", "Target host name from config file (mandatory for VM)")
```
mbulatova: Set empty default value here

ifilippov:
I suppose that we should change some names of files, however it is in separate commits.

I approve these changes if all scenarios were really tested and confirmed to work.

This include "make testing" from root directory, testing by "main" test system, building and deploying by "make" from root, clean by different clean variants. Also it shouldn't ruin other builds, e.g. "examples". Tests executable should be ignored by git.

dkolistratova:
@ifilippov,  do we really need these tests in root "make testing"? cause some tests require even 7 cores and travis machines may fail with "not enough cores" error. It now can be run from "test/stability" folder, should I really add to root makefile or leave as it is?

PR #206, opened 2018-02-09
Updated test system and vagrant script for automatic testing.

gshimansky:
- Added docker server setup and configuration to VMs config script.
- Removed docker port from launching in Makefiles because now it
  should be specified in YANFF_HOSTS.
- Moved test framework to its own subdirectory.
- Some cosmetic changes in ignore files.
- Changed test framework command line. Added hosts command line
  override.
- Switched vagrant script VMs to Fedora leaving Ubuntu as an option.
- Disabled FSGSBASE and F16C instructions to allow working in VM.
- Added verbose message after vagrant provisioning.
- In Makefiles added user name to docker images names to allow
  simultaneous work in multiuser mode.
- In test framework added username to images name. Added reaction to
  console interrupt with Ctrl-C.
- Added final statistics after tests finish.
- In Makefile set generated file dependency from types.go


ifilippov:
Approved, if you can prove that all these features were tested:

"make testing" from root directory
testing by "main" test system
building and deploying by "make" from root
clean by different clean variants
shouldn't ruin other builds, e.g. "examples"
tests executable should be ignored by git

Or at least those of them which were before this pull request.

PR #207, opened 2018-02-13
Refactored tests:

dkolistratova:
* renamed to camel-case instead of underscores
* fixed stability tests to fail with go test if they fail (was only
printing fail before)
* created citesting target to test with travis and testing target with
full stability tests


ifilippov:
Approved

PR #211, opened 2018-02-15
Updates to scripts and testing to run automated tests

gshimansky:
- Changed YANFF_HOSTS format to comma separated list host1:port,host2:port.
- Removed host-name and docker-port from test config files. Now they
  have to be specified on command line.
- Added environment variables with DPDK and Pktgen archive
  locations. Now it is possible to download them once and use file://
  URLs instead of downloading them every time.
- Made yanff-base image with username prefix too. This allows parallel
  work for multiple users on the same host.
- Fixed renaming artefacts.
- Fixed dependencies to build NFF-Go.
- Added optional docker image prefix.
- Added non-zero return status if some tests failed.

### DIFF
--- a/dpdk/Dockerfile
+++ b/dpdk/Dockerfile
@@ -1,8 +1,9 @@
-# Copyright 2017 Intel Corporation. 
+# Copyright 2017 Intel Corporation.
 # Use of this source code is governed by a BSD-style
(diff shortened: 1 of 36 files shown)


ifilippov:
Approved

PR #217, opened 2018-02-22
Fixed stability tests:

dkolistratova:
* refactored cksum test, removen unnecessary icmp packets from it, made
it skip unknown packets instead of fail, made it fail when no packets
received instead of pass
* fixed all tests to return error if fail was in statistics composition.

### DIFF
--- a/packet/checksum.go
+++ b/packet/checksum.go
@@ -22,6 +22,8 @@ func (packet *Packet) SetHWCksumOLFlags() {
 			low.SetTXIPv4TCPOLFlags(packet.CMbuf, EtherLen, IPv4MinLen)
 		} else if udp != nil {
 			low.SetTXIPv4UDPOLFlags(packet.CMbuf, EtherLen, IPv4MinLen)
+		} else {
+			low.SetTXIPv4OLFlags(packet.CMbuf, EtherLen, IPv4MinLen)
 		}
 	} else if ipv6 != nil {
 		tcp, udp, _ := packet.ParseAllKnownL4ForIPv6()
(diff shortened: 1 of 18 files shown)


packet/checksum.go
```diff
@@ -118,11 +120,16 @@ func SetPseudoHdrChecksum(p *Packet) {
...
 	if ipv4 != nil {
 		p.GetIPv4().HdrChecksum = 0
-		tcp, udp, _ := p.ParseAllKnownL4ForIPv4()
+		tcp, udp, icmp := p.ParseAllKnownL4ForIPv4()
 		if tcp != nil {
 			p.GetTCPForIPv4().Cksum = SwapBytesUint16(CalculatePseudoHdrIPv4TCPCksum(p.GetIPv4()))
 		} else if udp != nil {
 			p.GetUDPForIPv4().DgramCksum = SwapBytesUint16(CalculatePseudoHdrIPv4UDPCksum(p.GetIPv4(), p.GetUDPForIPv4()))
+		} else if icmp != nil {
+			// calculate full software checksum for icmp, cause there is no
```
mbulatova: Function SetPseudoHdrChecksum is not expected to calculate full checksums... @gshimansky what do you think?
I think it may be ok, but at least this should be stressed in func description

dkolistratova: I agree, but can't offer better variant. It is convenient to call SetPseudoHdrChecksum for packet and do nothing..there is no hard cksum for icmp and if we have possible tcp, udp, icmp packets with hwol mode, we need to parse packet ourselves, check whether it is icmp, calculate software cksum.

gshimansky: We can rename the function from `SetPseudoHdrChecksum` to something like `SetHWOffloadingHdrChecksum`, so it would be clear that this function is used when HW offloading is enabled. In case of ICMP this mode still requires software calculation, but for this we have a comprehensive comment here.

PR #221, opened 2018-02-23
Changed default DPDK log level in tests to 0

gshimansky:
- This is possibly a workaround for docker bugs with high output
  applications.
- Docker config via daemon.json made common for Ubuntu and Fedora
- Add 100 to number of failed tests for exit status. It allows quick
  diagnostics of error status in TeamCity.

### DIFF
--- a/examples/nat/main/nat.go
+++ b/examples/nat/main/nat.go
@@ -28,6 +28,7 @@ func main() {
 	configFile := flag.String("config", "config.json", "Specify config file name")
 	flag.BoolVar(&nat.CalculateChecksum, "csum", true, "Specify whether to calculate checksums in modified packets")
 	flag.BoolVar(&nat.HWTXChecksum, "hwcsum", true, "Specify whether to use hardware offloading for checksums calculation (requires -csum)")
+	dpdkLogLevel := flag.String("dpdk", "--log-level=0", "Passes an arbitrary argument to dpdk EAL")
 	flag.Parse()
 
 	// Set up reaction to SIGINT (Ctrl-C)
@@ -41,6 +42,7 @@ func main() {
 	nffgoconfig := flow.Config{
 		CPUList:      *cores,
 		HWTXChecksum: nat.HWTXChecksum,
+		DPDKArgs: []string{ *dpdkLogLevel },
 	}
 
 	CheckFatal(flow.SystemInit(&nffgoconfig))
--- a/test/framework/testsuite.go
+++ b/test/framework/testsuite.go
@@ -200,7 +200,12 @@ func (config *TestsuiteConfig) RunAllTests(logdir string) int {
 	LogInfo("TESTS EXECUTED:", totalTests)
 	LogInfo("PASSED:", passedTests)
 	LogInfo("FAILED:", failedTests)
-	return failedTests
+
+	if failedTests > 0 {
+		return 100 + failedTests
+	} else {
+		return 0
+	}
 }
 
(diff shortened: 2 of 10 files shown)


examples/nat/main/nat.go
```diff
@@ -28,6 +28,7 @@ func main() {
 	configFile := flag.String("config", "config.json", "Specify config file name")
 	flag.BoolVar(&nat.CalculateChecksum, "csum", true, "Specify whether to calculate checksums in modified packets")
 	flag.BoolVar(&nat.HWTXChecksum, "hwcsum", true, "Specify whether to use hardware offloading for checksums calculation (requires -csum)")
+	dpdkLogLevel := flag.String("dpdk", "--log-level=0", "Passes an arbitrary argument to dpdk EAL")
```
aregm: Better to write "specific" instead of "arbitrary", because that is what you do, you pass a specific argument to EAL.

test/framework/testsuite.go
```diff
@@ -200,7 +200,12 @@ func (config *TestsuiteConfig) RunAllTests(logdir string) int {
 	LogInfo("TESTS EXECUTED:", totalTests)
 	LogInfo("PASSED:", passedTests)
 	LogInfo("FAILED:", failedTests)
-	return failedTests
+
+	if failedTests > 0 {
+		return 100 + failedTests
```
aregm: Why 100 + tests? Is this for CI/CD?

vagrant/scripts.sh
```diff
@@ -147,20 +147,19 @@ setupdocker ()
...
-    sudo sh -c 'cat <<EOF > /etc/docker/daemon.json
-{
-    "hosts": ["unix:///var/run/docker.sock", "tcp://0.0.0.0:2375", "fd://"]
-}
-EOF'
     elif [ $DISTRO == Fedora ]; then
         sudo dnf config-manager --add-repo https://download.docker.com/linux/fedora/docker-ce.repo
         sudo dnf -y install docker-ce
         sudo gpasswd -a vagrant docker
         sudo firewall-cmd --permanent --add-port=2375/tcp
```
aregm: If there is a possibility to parametrize port, will be good.

examples/nat/main/nat.go
gshimansky: It is possible to pass a single arbitrary argument to DPDK EAL. The default value "--log-level=0" is specific, but it is possible to change it to anything else, e.g. not related to log level.

test/framework/testsuite.go
gshimansky: Yes, PR comment explains it. It is to easily distinguish common build/run errors with codes 1-10 from number of failed tests just by looking at the error code.

examples/nat/main/nat.go
aregm: Yes, but that is what specific means. Not arbitrary.

vagrant/scripts.sh
gshimansky: We use value 2375 in many other places, e.g. test framework configuration files to run tests. This action is done just once on a target machine, so if it is necessary to change docker daemon port, the user might as well edit this file once.

examples/nat/main/nat.go
gshimansky: Help message explains what this command line switch does. It allows user to pass an arbitrary switch string DPDK:

  -dpdk string
        Passes an arbitrary argument to dpdk EAL (default "--log-level=0")

PR #223, opened 2018-03-01
Support VLAN-tagged packets in testCksum

mbulatova:
- Add ParseDataCheckVLAN function
- ParseL4ForIPv4, ParseL4ForIPv6 work correctly with
VLAN-tagged packets

### DIFF
--- a/packet/checksum.go
+++ b/packet/checksum.go
@@ -14,23 +14,27 @@ import (
 
 // SetHWCksumOLFlags sets hardware offloading flags to packet
 func (packet *Packet) SetHWCksumOLFlags() {
-	ipv4, ipv6, _ := packet.ParseAllKnownL3()
+	ipv4, ipv6, _ := packet.ParseAllKnownL3CheckVLAN()
+	l2len := uint32(EtherLen)
+	if SwapBytesUint16(packet.Ether.EtherType) == VLANNumber {
+		l2len += VLANLen
+	}
 	if ipv4 != nil {
-		packet.GetIPv4().HdrChecksum = 0
+		packet.GetIPv4NoCheck().HdrChecksum = 0
 		tcp, udp, _ := packet.ParseAllKnownL4ForIPv4()
 		if tcp != nil {
-			low.SetTXIPv4TCPOLFlags(packet.CMbuf, EtherLen, IPv4MinLen)
+			low.SetTXIPv4TCPOLFlags(packet.CMbuf, l2len, IPv4MinLen)
 		} else if udp != nil {
-			low.SetTXIPv4UDPOLFlags(packet.CMbuf, EtherLen, IPv4MinLen)
+			low.SetTXIPv4UDPOLFlags(packet.CMbuf, l2len, IPv4MinLen)
 		} else {
-			low.SetTXIPv4OLFlags(packet.CMbuf, EtherLen, IPv4MinLen)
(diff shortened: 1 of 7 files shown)


packet/checksum.go
```diff
@@ -117,26 +121,26 @@ func CalculatePseudoHdrIPv6UDPCksum(hdr *IPv6Hdr, udp *UDPHdr) uint16 {
...
 		}
 	} else if ipv6 != nil {
 		tcp, udp, _ := p.ParseAllKnownL4ForIPv6()
 		if tcp != nil {
-			p.GetTCPForIPv6().Cksum = SwapBytesUint16(CalculatePseudoHdrIPv6TCPCksum(p.GetIPv6()))
+			p.GetTCPForIPv6().Cksum = SwapBytesUint16(CalculatePseudoHdrIPv6TCPCksum(ipv6))
 		} else if udp != nil {
-			p.GetUDPForIPv6().DgramCksum = SwapBytesUint16(CalculatePseudoHdrIPv6UDPCksum(p.GetIPv6(), p.GetUDPForIPv6()))
+			p.GetUDPForIPv6().DgramCksum = SwapBytesUint16(CalculatePseudoHdrIPv6UDPCksum(ipv6, udp))
 		}
```
dkolistratova: If we use -hwol flag in testScenario 2  and send icmp ipv6 packet, 2nd part will try to calculate hardware cksum and will fail, maybe it makes sense to include software variant here as for ipv4?

test/framework/main/stability.json
```diff
@@ -221,6 +221,90 @@
...
+        {
+            "name": "cksums_ipv4_tcp_vlan",
+            "test-time": 60000000000,
+            "test-type": "TestTypeScenario",
+            "test-apps": [
+                {
+                    "image-name": "nff-go-test-cksum",
+                    "app-type": "TestAppGo",
+                    "exec-cmd": [
+                        "./testCksum", "-testScenario=1", "-hwol", "-ipv4", "-tcp", "-vlan",  "-number=1000000", "-inport=0", "-outport=1"
```
dkolistratova: Do we need also test configuration with no hwol flag in 1st part (only in 2nd)?

mbulatova: We want checks HW vs SW Checksums in both part 1 and part 2.
I will add -hwol to 1st part in not-vlan tests

packet/checksum.go
mbulatova: Done

packet/vlan.go
```diff
@@ -139,3 +139,30 @@ func (packet *Packet) RemoveVLANTag() bool {
...
+
+// ParseDataCheckVLAN parses L3, L4 and fills the field packet.Data.
+// returns 0 in case of success and -1 in case of
+// failure to parse L3 or L4. VLAN presence is checked.
+func (packet *Packet) ParseDataCheckVLAN() int {
+	var pktTCP *TCPHdr
+	var pktUDP *UDPHdr
+	var pktICMP *ICMPHdr
+
+	pktIPv4, pktIPv6, _ := packet.ParseAllKnownL3CheckVLAN()
```
ifilippov: Why without ARP?

mbulatova: ParseData() does not parse ARP too
ARP packets has no user Data

ifilippov: Ok, get it.

ifilippov:
Approve

PR #227, opened 2018-03-02
Fix of core handling - can't use loop variable inside goroutines

ifilippov:


ifilippov:
Check failed due to DPDK site access, not because of commit.

PR #229, opened 2018-03-05
Merging scheduler into flow package

ifilippov:
Motivation:
Earlier we have only two types of functions& clonale and unclonable
Now we have generate function and will have receive function
in the future which have different scheduling heuristics. So
now it is very uncomfortable that scheduler doesnt know about
flow function parameters (this interface is known only inside flow package).

So we combine these two packets
Also we get rid off clonable, unclonable and other categories.
Now all flow functions form one list. And they are processed
based on their unique type.

The main functional change is that fastGenerate won't correct its
speed when scheduler is switched off. (Previously it doesn't use clones,
but correct speed).


ifilippov:
These changes aren't fully tested yet, however I want principal agreement that we want this refactor.

ifilippov:
Tested

PR #234, opened 2018-03-07
NAT performance tests

gshimansky:


ifilippov:
Please, integrate this into current tests. Remove NAT from perf.json or point the differences.

gshimansky:
These tests are different from perf.json because these tests require different network ports configuration before running. It is necessary to use linux kernel driver on client and server and assign IP addresses to these ports. Tests in perf.json require all ports to use DPDK driver.
Also perf-nat.json and perf-iptables.json require different network configuration too for the machine which runs NAT.

ifilippov:
These tests test different sides of NAT or the same side?

gshimansky:
These tests run real applications: apache benchmark and web server through NAT, this allows us to compare results with Linux iptables NAT. Tests in perf.json use pktgen to create UDP traffic, but presumably pktgen is able to create higher load than real applications.

ifilippov:
Then approved

Issue #233, opened 2018-03-07
Change in the Go toolchain 1.10 breaks the build.


aregm:
This commit https://github.com/golang/go/commit/7dea509703eb5ad66a35628b12a678110fbb1f72 breaks the build.

Quasilyte:
Build fails on all Go1.4+ versions with `-a` flag.
It does break on newer (1.10) even without `-a` build flag.

### Reason
NFF `make`-based Go build depends on the proper linkage (see below).
Go build may call linker (it does linking through `gcc` on linux)
on several object files. The exact combination depends on the
staleness and dependency resolving algorithms.
Those are changed in Go 1.10 significantly, hence build stopped to work.

NFF is so linkage-dependent due to `-Wl,--whole-archive` flag usage.
`go build` executes `gcc` with multiple copies of `CGO_LDFLAGS` (see `/mk/include.mk`),
and this leads to duplicate symbols:
```bash
# This is simplified command executed by cmd/link:
gcc obj1.o obj2.o \
  -Ldpdkpath \
  -Wl,--whole-archive -lrte_ring -lrte_mbuf -Wl,--no-whole-archive \
  -Ldpdkpath \
  -Wl,--whole-archive -lrte_ring -lrte_mbuf -Wl,--no-whole-archive
```
The main problem here is `-Wl,--whole-archive` around libraries.
Either there should be only one unique group inside them,
or it should not be used at all.

This simple example reproduces problem without NFF-specific stuff:
```bash
# Trivial program:
echo 'void main(int argc, char** argv) {}' > foo.c

# a) This builds OK:
gcc -o foo foo.c -lm -lm

# b) This causes multiple definitions errors:
gcc -o foo foo.c \
  -Wl,--whole-archive -lm -Wl,--no-whole-archive \
  -Wl,--whole-archive -lm -Wl,--no-whole-archive
```

(a) build successfully even though `-lm` passed twice (it could be `.a` file or `.o` - doesn't matter).
The only difference in (b) is both `-lm` wrapped inside `-Wl,--whole-archive`. This is exactly the case with NFF builds on Go1.10.

### Solution

The simplest solution is to avoid `-Wl,--whole-archive` if it's not needed.

> Note that it may require clean build. It's safer to run `go build` with `-a` flag during the first build or
> making other precautions to avoid cached artifacts usage.

The alternative solution is to re-structure build so that only one object file
depends on the dpdk directly. This can be hard to do in terms of `go build`,
[...]

PR #243, opened 2018-03-21
Fixed build to work with Go 1.10

gshimansky:
Fix for bug #233 
For best results use "go clean -cache" first if some files were built
with Go 1.10 previously.


Quasilyte:
I like the updated version.
LGTM.

PR #246, opened 2018-03-30
Example DPI application

mbulatova:
Read packets from file, allow or drop according to rules.
Support go regexp and Intel Hyperscan (https://01.org/hyperscan)
as matching engine.


.gitignore
```diff
@@ -1,3 +1,14 @@
...
+examples/demo
+examples/Firewall
+examples/Forwarding
+examples/dump
+examples/clonable_pcap_dumper
+examples/kni
+examples/copy
+examples/gopacket_parser_example
+examples/ErrorHandling
+test/stash/send_fixed_number
```
ifilippov: No, all of this is inside corresponding directories

.travis.yml
```diff
@@ -9,7 +9,7 @@ env:
     - MAKEFLAGS="-j2"
 
 dist:
-  - trusty
+  - xenial
```
ifilippov: Why?

examples/dpi/dpi.go
```diff
@@ -0,0 +1,374 @@
...
+	"flag"
+	"fmt"
+	"hash/fnv"
+	"io/ioutil"
+	"os"
+	"regexp"
+	"sync/atomic"
+	"time"
+
+	"github.com/flier/gohs/hyperscan"
```
ifilippov: Is everything good with dependencies here? Too small changes for new dependence? Maybe like gopacket?

examples/dpi/test-rules.json
```diff
@@ -0,0 +1,32 @@
+[
+    {
+        "name": "AllowFacebook",
```
ifilippov: AllowService1

examples/dpi/test-rules.json
```diff
@@ -0,0 +1,32 @@
...
+        "regexp": "Host: [a-z-]*.facebook.com",
+        "allow": true
+    },
+    {
+        "name": "AllowNetflix",
```
ifilippov: AllowService2

.travis.yml
mbulatova: xenial can install right gohs dependencies versions from repository (examples/dpi/setup-deps.sh)
Trusty has too old
(Unfortunately, no hyperscan in both, only in newer ubuntu)
There is option to run even more fresh os in Travis - in docker, i am trying it

examples/dpi/dpi.go
mbulatova: gohs requires pre-install some not-go dependencies
So I just added setup-deps.sh script, and use it CI on Ubuntu 16.04
And then make -C example/dpi

examples/dpi/Makefile
```diff
@@ -0,0 +1,9 @@
+# Copyright 2017 Intel Corporation.
```
TocarIP: 2018

examples/dpi/Makefile
```diff
@@ -0,0 +1,9 @@
...
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+PATH_TO_MK = ../../mk
+IMAGENAME = dpi
+EXECUTABLES = dpi
+
+include $(PATH_TO_MK)/leaf.mk
```
TocarIP: GitHub complains about lack of new line at the end of file

examples/dpi/dpi.go
```diff
@@ -0,0 +1,374 @@
...
+	for _, rule := range rules {
+		result := rule.Re.Match(data)
+		if !result {
+			continue
+		}
+		if rule.Allow {
+			accept = true
+		} else {
+			accept = false
+			atomic.AddUint64(numBlocked, 1)
```
TocarIP: So if 1 packet is disallowed by multiple rules we will increment numBlocked multiple times?
Looks like we can have numBlocked > numRead, which is strange

examples/dpi/dpi.go
```diff
@@ -0,0 +1,374 @@
...
+		pkt.ParseL4ForIPv6()
+	} else {
+		// Other protocols not supported
+		return totalNumFlows - 1
+	}
+	if ip4 != nil {
+		if ip4.NextProtoID != common.TCPNumber && ip4.NextProtoID != common.UDPNumber {
+			return totalNumFlows - 1
+		}
+		binary.Write(h, binary.BigEndian, ip4.NextProtoID)
```
TocarIP: Looks like ip4.NextProtoID is just an uint8, so big/little endian difference is confusing.
This IMHO looks better as:
h.write([]byte{ip4.NextProtoID})

examples/dpi/test-rules.json
```diff
@@ -0,0 +1,32 @@
+[
+    {
+        "name": "AllowService1",
+        "regexp": "Host: [a-z-]*.facebook.com",
```
dkolistratova: Service1 instead of facebook? The same below.

examples/dpi/dpi.go
```diff
@@ -0,0 +1,382 @@
...
+	CheckFatal(err)
+
+	CheckFatal(flow.SetSenderFile(outFlow, outfile))
+
+	go func() {
+		CheckFatal(flow.SystemStart())
+	}()
+
+	// Finish by timeout, as cannot verify if file reading finished
+	time.Sleep(timeout)
```
dkolistratova: Why it cant be handler with counter (nreads) and signal?

ifilippov:
Please rebase, than approve

mbulatova:
Rebased. Waiting for @aregm review

gshimansky:
Looks good to me although I have a question how this filter works if the pattern is fragmented between two or more packets. I didn't see packet defragmentation to search continuous contents.

examples/dpi/dpi.go
```diff
@@ -0,0 +1,382 @@
...
+	// Clones has the same id
+	id := lc.handlerId
+	newlc.handlerId = id
+	newlc.allowedCounterPtr = &allowedPktsCount[id]
+	newlc.readCounterPtr = &readPktsCount[id]
+	newlc.blockedCounterPtr = &blockedPktsCount[id]
+	return newlc
+}
+
+func (lc localCounters) Delete() {
```
aregm: Why you need this function if it is empty?

examples/dpi/dpi.go
```diff
@@ -0,0 +1,382 @@
...
+	atomic.AddUint64(numRead, 1)
+	data, err := extractData(pkt)
+	if err != nil {
+		fmt.Println("WARNING:", err, ", drop packet")
+		return false
+	}
+
+	accept := false
+	block := false
+	for _, rule := range rules {
```
aregm: Global rules is bad.

examples/dpi/dpi.go
```diff
@@ -0,0 +1,382 @@
...
+		atomic.AddUint64(numAllowed, 1)
+	}
+	if !accept && block {
+		atomic.AddUint64(numBlocked, 1)
+	}
+	return accept
+}
+
+func splitBy5Tuple(pkt *packet.Packet, context flow.UserContext) uint {
+	h := fnv.New64a()
```
aregm: Hmm? What is this hash for? Give some better name.

examples/dpi/dpi.go
```diff
@@ -0,0 +1,382 @@
...
+		buf := new(bytes.Buffer)
+		CheckFatal(binary.Write(buf, binary.LittleEndian, ip4.SrcAddr))
+		h.Write(buf.Bytes())
+		CheckFatal(binary.Write(buf, binary.LittleEndian, ip4.DstAddr))
+		h.Write(buf.Bytes())
+	} else if ip6 != nil {
+		if ip6.Proto != common.TCPNumber && ip6.Proto != common.UDPNumber {
+			return totalNumFlows - 1
+		}
+		binary.Write(h, binary.BigEndian, ip6.Proto)
```
aregm: This write cannot fail but the one on lines 242 and 244 can?

examples/dpi/dpi.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+
+	binary.Write(h, binary.BigEndian, pkt.GetTCPNoCheck().SrcPort)
+	binary.Write(h, binary.BigEndian, pkt.GetTCPNoCheck().DstPort)
+
+	hash := uint(h.Sum64())
+	return hash % numFlows
+}
+
+func extractData(pkt *packet.Packet) ([]byte, error) {
```
aregm: This function should be called getPacketPayload and be part of common module.

examples/dpi/dpi.go
```diff
@@ -0,0 +1,382 @@
...
+	ok := pkt.ParseData()
+	if ok == -1 {
+		return nil, fmt.Errorf("cannot extract packet data")
+	}
+
+	hdrsLen := uintptr(pkt.Data) - uintptr(pktStartAddr)
+	return pktBytes[hdrsLen:], nil
+}
+
+func getRulesFromFile(filename string) ([]rule, error) {
```
aregm: We have this function already in the examples. make it common.

examples/dpi/dpi.go
```diff
@@ -0,0 +1,382 @@
...
+	if err := json.Unmarshal(f, &rules); err != nil {
+		return nil, err
+	}
+	for i := 0; i < len(rules); i++ {
+		rules[i].Re = regexp.MustCompile(rules[i].Regexp)
+	}
+	return rules, nil
+}
+
+type rule struct {
```
aregm: Split example into several files. Create separate directory. Create package rules which will be responsible only for rules handling. Move the common functions to common and left only the logic to the main package.

examples/dpi/dpi.go
```diff
@@ -0,0 +1,382 @@
...
+			os.Exit(-1)
+		}
+		p.Id = k
+		patterns = append(patterns, p)
+	}
+	return
+}
+
+// Setup Hyperscan DBs and scratches
+func setupHyperscan(rules []rule) {
```
aregm: This should be part of rules or pattern package

examples/dpi/setup-deps.sh
```diff
@@ -0,0 +1,15 @@
...
+git clone https://github.com/intel/hyperscan.git
+cd hyperscan
+cmake . && make -j6
+
+export PKG_CONFIG_PATH=$HOME/hyperscan
+export LD_LIBRARY_PATH=$HOME/hyperscan/lib
+export CGO_LDFLAGS="-lstdc++ -L$HOME/hyperscan/lib -lhs"
+export CGO_CFLAGS="-I$HOME/hyperscan/src"
+
+go get -v github.com/flier/gohs/hyperscan
```
aregm: \n here

examples/dpi/dpi.go
mbulatova: To satisfy flow.UserContext interface

aregm: Ok

examples/dpi/dpi.go
mbulatova: This is just hash over 5-tuple, used to split same packets to same flow

examples/dpi/dpi.go
mbulatova: I think it is ok because rules accessed only for reading

examples/dpi/dpi.go
mbulatova: Fixed

examples/dpi/dpi.go
mbulatova: This is different one. Renamed into getPatternsFromFile and rule<->pattern to avoid confusion

examples/dpi/dpi.go
mbulatova: Suppose it should go into 'packet', with test and in separate commit
@ifilippov, is it ok?

PR #253, opened 2018-04-03
Fix test speeds

ifilippov:
1) Fix computation of final numbers. Now sent and received are more accurate
2) Tiny refactoring
3) Gofmt


test/stability/testSplit/testSplit.go
```diff
@@ -363,6 +363,10 @@ func checkInputFlow2(pkt *packet.Packet, context flow.UserContext) {
 		return
```
mbulatova: Please join in one condition

test/stability/testSplit/testSplit.go
```diff
@@ -332,6 +331,10 @@ func checkInputFlow1(pkt *packet.Packet, context flow.UserContext) {
...
```
mbulatova: One condition

test/stability/testMerge/testMerge.go
```diff
@@ -230,7 +227,7 @@ func composeStatistics() error {
 	// Test is passed, if p1 and p2 do not differ too much: |p1-p2| < 4%
 	// and enough packets received back
 	if atomic.LoadInt32(&passed) != 0 &&
-		(p1-p2 < 4 || p2-p1 < 4) && received*100/sent > passedLimit {
+		(p1-p2 < 4 && p1-p2 > -4) && received*100/sent > passedLimit {
```
dkolistratova: I think it was more simple and obvious before...

ifilippov: Maybe, unfortunately it was mathematical incorrect, because "|x|<y" is logical AND, not OR

test/stability/testSeparate/testSeparate.go
```diff
@@ -55,6 +55,7 @@ var (
 	recvCount2 uint64
 
 	count         uint64
+	countRes      uint64
```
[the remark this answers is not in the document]
ifilippov: We should change count anytime, because we generate packet's difference from it. And countRes is count for time more than T.

(if you don't change count you will have a lot of packets  in the very beginning of T at receive side which will be the same. They will be calculated as if separate doesn't work)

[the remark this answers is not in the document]
dkolistratova: We skip packets in the very beginning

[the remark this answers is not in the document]
ifilippov: Moment T is achieved simultaneously for send and for receive. If you send the same packets before T, receive will receive the same packets by "T + latency" time. And it will think that separate doesn't work, because all the packets are the same. It means that you need to change packets even before "T". But calculate them only after "T".

PR #255, opened 2018-04-04
Important hotfix for master after switch to DPDK 18

gshimansky:
Fixed HW checksums after switching to DPDK 18

In new DPDK version rte_mbuf struct received a new union field which
Go language marks as anonymous fields because it cannot address them
in any other way. This causes anonX field numbering to shift and
anonymous field's number which stores packet lengths information had
to be corrected.

### DIFF
--- a/low/low.go
+++ b/low/low.go
@@ -157,19 +157,24 @@ func TrimMbuf(m *Mbuf, length uint) bool {
 	return true
 }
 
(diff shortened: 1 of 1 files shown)


ifilippov:
Not for master

ifilippov:
Add a comment like "anon = real_name"

ifilippov:
Approved (maybe change commit name?)

PR #256, opened 2018-04-04
Combined clonable flow functions to one "Segment" flow function

ifilippov:
Changes short:
1) Sequential segments of functions handle, split, separate and partition are now
- invocated at one core
- clonned simultaneously
- don't use any rings and mbuf inside them
2) Vector functions use mask instead of packets regrouping
3) Partition is now clonnable function

Changes long:
Flow graph is now divided into four parts.
1) "Get" functions: Receive, Receive from KNI, Receive from File, Generate, Fast generate
2) "Copy" function: Copy
3) "Segment" functions: Handle, Separate, Split, Partition
4) "Set" functions: Send, Send to KNI, Send to file, Stop
1 - is always the first.
2 - can be after 1 or 3
3 - can be after 1 or 2
4 - can be after 1, 2 or any function from 3 group
Merge function interrupts sequential segment and can be placed wherever user wants
Transitions between Scalar and vector code also interrupts segment

--------------------

Performance results for our four main parameters are the following:

Mb/s, no scheduler - drop significantly. Can't compare this directly, because all handling is at one core now. Reduction is expected.

Cores number, no scheduler - drop by 1 at non sequential tests (handle + partition were combined). Dropped significantly at sequential tests (sequential handles + final partition combined)

Mb/s, scheduler - increased by 5.85% in geomean. Lot's more increasing in sequential tests including NAT. Performance of lighter test not significantly drop (need to be investigated)

Cores number, scheduler - drop by 6.06% geomean. Significantly dropped at sequential tests. Slightly increased at NAT.

Latency dropped because our tests checked it without cloning. This means that earlier we have four sequential cores with handles per test which which worked as pipeline. Now we have only one core and latency increases.

### DIFF
--- a/examples/tutorial/step10.go
+++ b/examples/tutorial/step10.go
@@ -9,6 +9,8 @@ import "github.com/intel-go/nff-go/packet"
 
 var rulesp unsafe.Pointer
 
+const vecSize = 32
+
 func main() {
 	config := flow.Config{}
 	checkFatal(flow.SystemInit(&config))
@@ -38,15 +40,17 @@ func mySplitter(cur *packet.Packet, ctx flow.UserContext) uint {
 	return cur.L3ACLPort(localL3Rules)
 }
 
-func myHandler(curV []*packet.Packet, num uint, ctx flow.UserContext) {
-	for i := uint(0); i < num; i++ {
-		cur := curV[i]
-		cur.EncapsulateHead(common.EtherLen, common.IPv4MinLen)
-		cur.ParseL3()
-		cur.GetIPv4NoCheck().SrcAddr = packet.BytesToIPv4(111, 22, 3, 0)
-		cur.GetIPv4NoCheck().DstAddr = packet.BytesToIPv4(3, 22, 111, 0)
-		cur.GetIPv4NoCheck().VersionIhl = 0x45
-		cur.GetIPv4NoCheck().NextProtoID = 0x04
+func myHandler(curV []*packet.Packet, mask *[vecSize]bool, ctx flow.UserContext) {
+	for i := uint(0); i < 32; i++ {
+		if (*mask)[i] == true {
+			cur := curV[i]
+			cur.EncapsulateHead(common.EtherLen, common.IPv4MinLen)
+			cur.ParseL3()
+			cur.GetIPv4NoCheck().SrcAddr = packet.BytesToIPv4(111, 22, 3, 0)
+			cur.GetIPv4NoCheck().DstAddr = packet.BytesToIPv4(3, 22, 111, 0)
+			cur.GetIPv4NoCheck().VersionIhl = 0x45
+			cur.GetIPv4NoCheck().NextProtoID = 0x04
+		}
 	}
 }
 
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -46,11 +46,51 @@ var openFlowsNumber = uint32(0)
 var ringName = 1
 var createdPorts []port
 var schedState *scheduler
+var vEach [10][burstSize]uint8
+
+type processSegment struct {
+	out      []*low.Ring
+	contexts []UserContext
+	stype    uint8
+}
 
 // Flow is an abstraction for connecting flow functions with each other.
 // Flow shouldn't be understood in any way beyond this.
 type Flow struct {
-	current *low.Ring
+	current  *low.Ring
+	segment  *processSegment
+	previous **Func
+}
+
+type partitionCtx struct {
+	currentAnswer       uint8
+	currentCompare      uint64
+	currentPacketNumber uint64
+	N                   uint64
+	M                   uint64
+}
+
+func (c partitionCtx) Copy() interface{} {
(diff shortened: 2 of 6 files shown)


examples/tutorial/step10.go
```diff
@@ -38,15 +38,17 @@ func mySplitter(cur *packet.Packet, ctx flow.UserContext) uint {
...
-func myHandler(curV []*packet.Packet, num uint, ctx flow.UserContext) {
-	for i := uint(0); i < num; i++ {
-		cur := curV[i]
-		cur.EncapsulateHead(common.EtherLen, common.IPv4MinLen)
-		cur.ParseL3()
-		cur.GetIPv4NoCheck().SrcAddr = packet.BytesToIPv4(111, 22, 3, 0)
-		cur.GetIPv4NoCheck().DstAddr = packet.BytesToIPv4(3, 22, 111, 0)
-		cur.GetIPv4NoCheck().VersionIhl = 0x45
-		cur.GetIPv4NoCheck().NextProtoID = 0x04
+func myHandler(curV []*packet.Packet, mask *[32]bool, ctx flow.UserContext) {
```
TocarIP: Replace 32 with some constant?
E. g.
const vecSize = 32
...
func myHandler(....mask *[vecSize]bool...
...
for .. i < vecSize

flow/flow.go
```diff
@@ -46,11 +46,55 @@ var openFlowsNumber = uint32(0)
...
+type partitionCtx struct {
+	currentAnswer       int
+	currentCompare      uint64
+	currentPacketNumber uint64
+	N                   uint64
+	M                   uint64
+}
+
+func (c partitionCtx) Copy() interface{} {
+	n := new(partitionCtx)
```
TocarIP: Do you still hate struct literals?
return partitionCtx{N:c.N,M;c.M,currentCompare:c.N} is just 1 line...

flow/flow.go
```diff
@@ -46,11 +46,55 @@ var openFlowsNumber = uint32(0)
...
+
+func (c partitionCtx) Delete() {
+}
+
+type Func struct {
+	sHandleFunction   HandleFunction
+	sSeparateFunction SeparateFunction
+	sSplitFunction    SplitFunction
+	sFunc             func(*packet.Packet, *Func, UserContext) int
+        vHandleFunction   VectorHandleFunction
```
TocarIP: Formatting looks weird on github

flow/flow.go
```diff
@@ -591,229 +608,351 @@ func SetSender(IN *Flow, portId uint8) error {
...
+			if scalar { // Scalar code
+				for i := uint(0); i < n; i++ {
+                                       currentFunc := firstFunc
+					tempPacketAddr = packet.ExtractPacketAddr(InputMbufs[i])
+					tempPacket = packet.ToPacket(tempPacketAddr)
+					for {
+						nextIndex := currentFunc.sFunc(tempPacket, currentFunc, context[currentFunc.contextIndex])
+						if nextIndex < 0 {
+							// Previous func was constructSlice
+							index := nextIndex*-1 - 1
```
TocarIP: -nextIndex-1 ?

flow/flow.go
```diff
@@ -591,229 +608,351 @@ func SetSender(IN *Flow, portId uint8) error {
...
+	}
+	var currentSpeed uint64
+	tick := time.Tick(time.Duration(schedTime) * time.Millisecond)
+	var pause int
+	firstFunc := lp.firstFunc
+	// For scalar part
+	var tempPacket *packet.Packet
+	var tempPacketAddr uintptr
+	// For vector part
+	tempPackets := make([]*packet.Packet, burstSize)
```
TocarIP: Move those to use?

ifilippov: I should create slice above loop for performance reasons

flow/flow.go
```diff
@@ -46,11 +46,55 @@ var openFlowsNumber = uint32(0)
...
+
+func (c partitionCtx) Copy() interface{} {
+	n := new(partitionCtx)
+	n.N = c.N
+	n.M = c.M
+	n.currentCompare = c.N
+	return n
+}
+
+func (c partitionCtx) Delete() {
```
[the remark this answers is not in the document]
ifilippov: For interface acceptance

flow/flow.go
```diff
@@ -591,229 +608,351 @@ func SetSender(IN *Flow, portId uint8) error {
...
+		closeFlow(IN)
+	} else {
+		IN.previous = &f.next[0]
+	}
+	IN.line.contexts = append(IN.line.contexts, context)
+	f.contextIndex = len(IN.line.contexts) - 1
+	return nil
+}
+
+func lineProcess(parameters interface{}, stopper chan int, report chan uint64, context []UserContext) {
```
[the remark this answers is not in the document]
ifilippov: All parameters are interfaces because they are different for different Flow Functions

ifilippov:
Latency dropped because our tests checked it without cloning. This means that earlier we have four sequential cores with handles per test which which worked as pipeline. Now we have only one core and latency increases.

PR #274, opened 2018-04-12
Extracted CheckFatal function to flow.go, plus fixed a bug in MergeTest.

dkolistratova:


test/stability/testMerge/testMerge.go
```diff
@@ -195,7 +195,7 @@ func executeTest(configFile, target string, testScenario uint) error {
 		testDoneEvent.Wait()
 		testDoneEvent.L.Unlock()
 
-		composeStatistics()
+		return composeStatistics()
```
ifilippov: Good catch! It is really working now?

ifilippov:
Some problems with checks (not all "os" and "fmt" imports were removed). Besides this - Approved.

dkolistratova:
@ifilippov  seems like make does not build...cause I checked make all in main dir, test dir, examples dir. It it normal?

test/stability/testMerge/testMerge.go
dkolistratova: It worked, a bug is that when test actually fails, nil error is returned instead of real error and gotest says all ok, when it actually was not.

ifilippov:
Merged due to following pull requests dependencies

PR #277, opened 2018-04-13
Perf tests renaming

ifilippov:
Remove sendback from performance
Rename perf-iptables to perf-nat-linux


gshimansky:
Let's keep old name of sendbank.go file. It looks more explanative to me.

PR #281, opened 2018-04-16
Cleanup: move files from test/stash to examples

mbulatova:


ifilippov:
Can you please place .pcap file with it's reader in separate directory inside examples. (Can you also place demo.go with its .conf and .orig files also in separate demo directory and Forwarding with Firewall also to two separate directories)?

PR #282, opened 2018-04-17
Added functions SystemStop and SystemReset

ifilippov:
SystemStop - stopping all framework
SystemReset - stopping framework and releasing DPDK (DPDK limitations for now)


gshimansky:
Approved changes in scheduler. Please do changes in testDivide in a separate patch.

PR #284, opened 2018-04-18
Add packet fields changing to our flow function test. Remove testRese…

ifilippov:
…nd as bad and duplicated


test/stability/testDivide/testDivide.go
```diff
@@ -106,10 +109,10 @@ func main() {
 	dpdkLogLevel = *(flag.String("dpdk", "--log-level=0", "Passes an arbitrary argument to dpdk EAL"))
 	flag.Parse()
 	if err := initDPDK; err != nil {
-		fmt.Printf("fail: %+v\n", err)
+		println("fail: %+v\n", err)
```
mbulatova: printf

test/stability/testDivide/testDivide.go
```diff
@@ -106,10 +109,10 @@ func main() {
...
 	}
 	if err := executeTest(*configFile, *target, testScenario, testType); err != nil {
-		fmt.Printf("fail: %+v\n", err)
+		println("fail: %+v\n", err)
```
mbulatova: printf

test/stability/testDivide/testDivide.go
```diff
@@ -106,10 +109,10 @@ func main() {
 	dpdkLogLevel = *(flag.String("dpdk", "--log-level=0", "Passes an arbitrary argument to dpdk EAL"))
 	flag.Parse()
 	if err := initDPDK; err != nil {
```
dkolistratova: wow did you test if it should not work?! initDPDK is not called at all, always err != nil.

mbulatova: initDPDK()

test/stability/testDivide/testDivide.go
```diff
@@ -129,6 +132,13 @@ func executeTest(configFile, target string, testScenario uint, testType uint) er
 		return errors.New("testScenario should be in interval [0, 3]")
 	}
 	gTestType = testType
+	countRes = 0
```
dkolistratova: move initialization to place where vars declared

test/stability/testDivide/testDivide.go
```diff
@@ -235,13 +246,15 @@ func executeTest(configFile, target string, testScenario uint, testType uint) er
 func setParameters(testScenario uint) (err error) {
 	eps = 3
 	passedLimit = 85
+	shift = 0
```
dkolistratova: too many magics, replace with names. 

By the way, logic of check packets is not obvious now, needs comments here and in place where flows are checked.

test/stability/testDivide/testDivide.go
```diff
@@ -235,13 +246,15 @@ func executeTest(configFile, target string, testScenario uint, testType uint) er
...
 	switch gTestType {
 	case separate:
 		if testScenario != generatePart {
 			l3Rules, err = packet.GetL3ACLFromORIG("test-separate-l3rules.conf")
 		}
 		// Test expects to receive 33% of packets on 0 port and 66% on 1 port
 		lessPercent = 33
+		shift = 1
```
mbulatova: Why no shift in split, partition, pcopy?

test/stability/testDivide/testDivide.go
```diff
@@ -259,15 +272,17 @@ func setParameters(testScenario uint) (err error) {
...
+		shift = 3
 	case dhandle:
 		if testScenario != generatePart {
 			l3Rules, err = packet.GetL3ACLFromORIG("test-handle-l3rules.conf")
 		}
 		// Test expects to receive 100% of packets on 0 port and 0% on 1 port (33% total)
 		lessPercent = 100
-		passedLimit = 28 // 85 * 0.33
+		passedLimit = 28       // 85 * 0.33
+		totalPackets = 3300000 // 10 000 000 * 0.33
```
dkolistratova: why do we change total packets number? what if user have set total packets to 100 and now you set it to 3300000? Magics are evil!!!

test/stability/testDivide/testDivide.go
```diff
@@ -394,93 +398,106 @@ func generatePacket(pkt *packet.Packet, context flow.UserContext) {
 		// Generate identical packets
 		udp.DstPort = packet.SwapBytesUint16(dstPort1)
 	}
-	ipv4.HdrChecksum = packet.SwapBytesUint16(packet.CalculateIPv4Checksum(ipv4))
-	udp.DgramCksum = packet.SwapBytesUint16(packet.CalculateIPv4UDPChecksum(ipv4, udp, pkt.Data))
-	fixMACAddrs(pkt, context)
-	atomic.AddUint64(&count, 1)
+
+	// Put a unique non-zero value
```
dkolistratova: you tell "put a unique", but in fact put always 2 below, comment conflicts with code, change it.

test/stability/testDivide/testDivide.go
```diff
@@ -394,93 +398,106 @@ func generatePacket(pkt *packet.Packet, context flow.UserContext) {
...
+	ipv4 := pkt.GetIPv4()
+	udp := pkt.GetUDPForIPv4()
+	recvIPv4Cksum := packet.SwapBytesUint16(packet.CalculateIPv4Checksum(ipv4))
+	recvUDPCksum := packet.SwapBytesUint16(packet.CalculateIPv4UDPChecksum(ipv4, udp, pkt.Data))
+	ptr := (*packetData)(pkt.Data)
+	ptr.F2 -= shift
+	if recvIPv4Cksum != ipv4.HdrChecksum || recvUDPCksum != udp.DgramCksum ||
+		ptr.F2 != 2 && ptr.F2 != 7 ||
+		ptr.F2 == 2 && ptr.F1 != 1 ||
+		gTestType != pcopy && ptr.F2 == 7 && checkSlice[ptr.F1] != 0 ||
```
dkolistratova: not obvious at all, add comment and change magics to constants, they are the same in more than one place, should be changed.

test/stability/testDivide/testDivide.go
```diff
@@ -394,93 +398,106 @@ func generatePacket(pkt *packet.Packet, context flow.UserContext) {
...
+}
+
 func l3Separator(pkt *packet.Packet, context flow.UserContext) bool {
-	// Return whether packet is accepted
+	addF2(pkt)
 	return pkt.L3ACLPermit(l3Rules)
 }
 
 func l3Splitter(currentPacket *packet.Packet, context flow.UserContext) uint {
-	// Return number of flow to which put this packet. Based on ACL rules.
```
dkolistratova: why did you remove comments here and above?

test/stability/testDivide/testDivide.go
ifilippov: Forget to git add a file - good catch, thank you!

test/stability/testDivide/testDivide.go
ifilippov: No. All tests are re-init now. This variables should be zero before each run.

test/stability/testDivide/testDivide.go
ifilippov: You want to replace shift = 2 with names? Maybe with shift = two?

test/stability/testDivide/testDivide.go
ifilippov: No. I place unique values below. It is in same logic.

test/stability/testDivide/testDivide.go
ifilippov: Good catch! - Thank you!

test/stability/testDivide/testDivide.go
dkolistratova: shiftForSplit, shiftDefault and so on.

ifilippov: And how to initialize them? "shiftForSplit" = 2 is magic number isn't it?

dkolistratova: yes, but when you meet 2 somewhere you understand what it is. And lower risks to forget during future changes.

PR #292, opened 2018-04-19
Change port type to uint16 in line with DPDK

ifilippov:
Plus options cleanup

After this cleanup it is very clear that we need to deal with global variables passing to handle functions. Something like contexts.

### DIFF
--- a/test/stability/testDivide/testDivide.go
+++ b/test/stability/testDivide/testDivide.go
@@ -70,10 +70,10 @@ var (
 	// During timeout packets are skipped and not counted
 	T = 10 * time.Second
 
-	outport1     uint
-	outport2     uint
-	inport1      uint
-	inport2      uint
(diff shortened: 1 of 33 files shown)


test/stability/testDivide/testDivide.go
```diff
@@ -151,7 +151,7 @@ func executeTest(configFile, target string, testScenario uint, testType uint) er
 	}
 
 	if testScenario == receivePart {
-		inputFlow, err := flow.SetReceiver(uint8(inport1))
+		inputFlow, err := flow.SetReceiver(uint16(inport1))
```
mbulatova: Maybe also convert to uint16 once, like do in other files?
inport1 = uint16(*flag.Uint(...))

ifilippov: OK, yes!

PR #298, opened 2018-04-23
Fixed error printing function, was not properly formatting messages.

dkolistratova:


ifilippov:
I suppose that fixing this should be around sprinln in LogFatal. Not arounf adding fmt to flow.go

ifilippov:
Approved

PR #300, opened 2018-04-24
Switched vagrant script to work with libvirt

gshimansky:
An attempt has been made to preserve functioning VirtualBox
provisioning, but it wasn't tested. Updated getports script and README
with instructions on how to create reference set of VMs using libvirt.

### DIFF
--- a/vagrant/.gitignore
+++ b/vagrant/.gitignore
@@ -1,2 +1,3 @@
 *.log
 .vagrant
+config.ssh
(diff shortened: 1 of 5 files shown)


ifilippov:
My propose is the following:
1) You spent lots of time from VirtualBox configure - it is pity to remove it.
2) You can check if VirtualBox works now
3) And if it isn't maybe we should keep to different files or even different directories?
I suppose it shouldn't be committed before these checks and decision.

gshimansky:
Tested with VirtualBox, in fact all that was required is to change MAC address format. The change works with VB now too.

PR #303, opened 2018-04-24
Add vector tests

ifilippov:


test/stability/testDivide/testDivide.go
```diff
@@ -313,6 +328,20 @@ func setMainTest(inFlow *flow.Flow) (*flow.Flow, *flow.Flow, error) {
 		err = flow.SetHandler(inFlow, l3Handler, nil)
 	case dhandle:
 		err = flow.SetHandlerDrop(inFlow, l3dHandler, nil)
+	case vseparate:
+		// Separate packet flow based on ACL.
+		flow2, err = flow.SetVectorSeparator(inFlow, vectorL3Separator, nil)
+	case vpartition:
+		// We need vector handler to switch to vector mode
```
dkolistratova: looks like a kludge...(not in test, but in realization) but if handle is broken, partition will be broken too.

ifilippov: Partition is not user defined function so it has a vector/scalar type as its surrounding to not ruin segments. All other test is scalar, so we should have at least one vector function to make partition vector.

Issue #308, opened 2018-04-26
panic: interface conversion: flow.UserContext is nil, not main.generatorParameters 


dkolistratova:
Generator does not work with user context. Reproducer:

```go
package main

import (
	"time"

	"github.com/intel-go/nff-go/flow"
	"github.com/intel-go/nff-go/packet"
)

var (
	payloadSize uint   = 16
	ipv4addr1 uint32 = 0x0100007f // 127.0.0.1
	ipv4addr2 uint32 = 0x05090980 // 128.9.9.5
)

func main() {
	config := flow.Config{
		DPDKArgs: []string{"--log-level=0"},
	}
	flow.CheckFatal(flow.SystemInit(&config))
	gp := generatorParameters{isGroup1:true}
	flow1 := flow.SetGenerator(generatePacketGroup, &gp)
	flow.CheckFatal(flow.SetStopper(flow1))
	var err error
	go func() {
		err = flow.SystemStart()
	}()
	flow.CheckFatal(err)
	time.Sleep(10 * time.Second)
}


type generatorParameters struct {
	isGroup1 bool
}

func (p generatorParameters) Copy() interface{} {
	return p
}

func (p generatorParameters) Delete() {
}

func generatePacketGroup(pkt *packet.Packet, context flow.UserContext) {
	p := context.(generatorParameters)
	isGroup1 := p.isGroup1
	if pkt == nil {
		panic("Failed to create new packet")
	}
	if packet.InitEmptyIPv4UDPPacket(pkt, payloadSize) == false {
		panic("Failed to init empty packet")
	}
	ipv4 := pkt.GetIPv4()
	if isGroup1 {
		ipv4.SrcAddr = ipv4addr1
	} else {
		ipv4.SrcAddr = ipv4addr2
	}
}
```

fails with panic: interface conversion: flow.UserContext is nil, not main.generatorParameters

ifilippov:
Done, close.

PR #309, opened 2018-04-26
Fix of generate flow function: user contexts

ifilippov:
Plus some internal renaming


ifilippov:
https://github.com/intel-go/nff-go/issues/308

PR #310, opened 2018-04-26
Fixed checksum tests. All 12 tests pass in VMs now.

gshimansky:


ifilippov:
I don't approve. What will be with all these new options without VMs? What will be with new speed without VM?

ifilippov:
I suppose that stability.json should stay for non-VM environments. We can test it nightly at our servers. And new changes should be in stability-VM for VM testing. With any workarounds and speed reductions.

PR #312, opened 2018-04-28
Fix flags parsing

mbulatova:
was impossible to reset default values


ifilippov:
Let's convert to uint16* in Flag and use as "*inport" later. Can we do this?

ifilippov:
Unfortunately we can't do it. Approve.

PR #318, opened 2018-05-02
Eliminate overriding recipe while make

ifilippov:
The main change is removing include.mk from intermediate.mk. This include leads to duplication in example directory, confusing warnings and debug messages because examples directory is considered both leaf and intermediate with both includes. When both of them include include.mk - it was duplication.


ifilippov:
Also I places check-pktgen checking everywhere. Before this fix it was only in packet - "make testing". However it is useful checking and we should have it before all "make testing" callings.

PR #324, opened 2018-05-03
Added default value for variable cause it was not set by gotest and failed.

dkolistratova:
Tests that can be run by gotest should have this initialization.


ifilippov:
How can you explain what has happened between last week and now? Please not merge before this explanation.

mbulatova:
@ifilippov, it was deleted by mistake in 4b3d81cae865e348ecf496b31d7af380c3946461

Issue #289, opened 2018-04-19
gotest testChecksum


ifilippov:
We have ~10 checksum tests in stability.json and only one in cksum_test.go. Is this right? You can use SystemStop now.

ifilippov:
It shouldn't hang after https://github.com/intel-go/nff-go/commit/3611beb9f55b3959e5d901f65725695b0560a761

PR #333, opened 2018-05-07
Modified go test for cksum

dkolistratova:
#289

### DIFF
--- a/test/stability/testCksum/checksum_test.go
+++ b/test/stability/testCksum/checksum_test.go
@@ -8,9 +8,34 @@ import (
 	"testing"
 )
 
+func TestInit(t *testing.T) {
+	if err := initTest(); err != nil {
+		t.Fatalf("fail to initialize with error: %+v\n", err)
+	}
+}
+
+// TODO unfortunately this test can't be executed as go test right now due to https://github.com/intel-go/nff-go/issues/301
+// If you want to test it via go test you should firstly comment FreeMempools function at flow/flow.go
+// You will have multiple mempools however test will pass.
 func TestChecksum(t *testing.T) {
-	if err := executeTest(0); err != nil {
-		t.Logf("fail: %+v\n", err)
-		t.Fail()
+	variants := []struct{ useUDP, useTCP, useICMP bool }{
+		{false, false, false},
+		{true, false, false},
+		{false, true, false},
+		{false, false, true},
+		{true, true, false}}
+	for _, useIPv4 := range []bool{true, false} {
+		for _, useIPv6 := range []bool{true, false} {
+			for _, useVLAN := range []bool{true, false} {
+				for _, v := range variants {
+					err := executeTest(gotest, useIPv4, useIPv6, v.useUDP, v.useTCP, v.useICMP, useVLAN)
+					resetState()
+					if err != nil {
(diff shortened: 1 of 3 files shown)


test/stability/testCksum/checksum_test.go
```diff
@@ -8,9 +8,36 @@ import (
...
+    }
+}
+// TODO unfortunately this test can't be executed as go test right now due to https://github.com/intel-go/nff-go/issues/301
+// If you want to test it via go test you should firstly comment FreeMempools function at flow/flow.go
+// You will have multiple mempools however test will pass.
 func TestChecksum(t *testing.T) {
-	if err := executeTest(0); err != nil {
-		t.Logf("fail: %+v\n", err)
-		t.Fail()
+	variants := []bool {true, false}
```
aregm: Too bizzare. I would go with test pattern table variants[][] bool {{false, false,false,false,false,false},} This will eliminate nested fors and check for presence of ICMP only with tcp/udp.

test/stability/testCksum/testCksum.go
```diff
@@ -100,6 +106,27 @@ func main() {
...
+	useTCP = shouldUseTCP
+	useUDP = shouldUseUDP
+	useVLAN = shouldUseVLAN
+	if hwol {
+		print("hardware cksums and ")
+	}
+	if useVLAN {
+		print("vlan tag and ")
+	}
 	if useIPv4 {
```
aregm: && ?

test/stability/testCksum/testCksum.go
```diff
@@ -136,35 +163,49 @@ func main() {
...
 
-	if hwol {
-		testCksumCommon.VirtualBoxWorkaround = *vbwo
-		testCksumCommon.KVMWorkaround = *kvmwo
-	}
-
-	if err := executeTest(testScenario); err != nil {
-		fmt.Printf("fail: %+v\n", err)
-	}
+func resetState() {
```
aregm: please run go fmt

test/stability/testCksum/checksum_test.go
```diff
@@ -8,9 +8,36 @@ import (
...
+					for _, useICMP := range variants {
+						if useICMP && (useTCP || useUDP) {
+							continue
+						}
+						for _, useVLAN := range variants {
+							err := executeTest(gotest, useIPv4, useIPv6, useUDP, useTCP, useICMP, useVLAN)
+							resetState()
+							if err != nil {
+								t.Logf("useIPv4=%t, useIPv6=%t, useUDP=%t, useTCP=%t, useICMP=%t, useVLAN=%t configuration failed with: %+v\n", useIPv4, useIPv6, useUDP, useTCP, useICMP, useVLAN, err)
+								t.Fail()
```
aregm: or failTest()?

dkolistratova: no, these functions are gotest specific, they should be used.

test/stability/testCksum/testCksum.go
dkolistratova: it is just for information which configuration is used. This style was used before, I just appended vlan to it.

test/stability/testCksum/testCksum.go
dkolistratova: done

ifilippov:
Approve (maybe we should exclude part of tests, however it is later).

PR #337, opened 2018-05-07
Small addition to scheduler heuristic

ifilippov:
We shouldn't clone anything if next ring is overloaded.
It increases pressure for our release system and
use unnecessary cores.

### DIFF
--- a/flow/scheduler.go
+++ b/flow/scheduler.go
@@ -347,12 +347,13 @@ func (scheduler *scheduler) schedule(schedTime uint) {
 			}
 			// Secondly we check adding clones. We can add one clone if:
 			// 1. scheduler is switched on
-			if scheduler.off == false {
+			// 2. ouput ring is quite empty
+			if scheduler.off == false && ff.checkOutputRingClonable(scheduler.maxPacketsToClone) {
 				switch ff.fType {
 				case segmentCopy:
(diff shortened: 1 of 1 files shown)


flow/scheduler.go
```diff
@@ -496,10 +498,28 @@ func (ff *flowFunction) checkInputRing() (n uint32) {
 	return n
 }
 
-func (ff *flowFunction) checkOutputRing() (n uint32) {
+func (ff *flowFunction) checkOutputRing(max uint32) bool {
```
aregm: This function should be renamed to isOutputRingClonable, the same for input ring

flow/scheduler.go
```diff
@@ -395,7 +396,8 @@ func (scheduler *scheduler) schedule(schedTime uint) {
 					if recC == -1 {
 						recC = low.CheckRSSPacketCount(ff.Parameters.(*receiveParameters).port)
 					}
-					if recC > 39 && ff.checkOutputRing() <= scheduler.maxPacketsToClone {
+					// 3. Number of packets in RSS is big enogh to cause overwrite (drop) problems
+					if recC > 39 {
```
aregm: Should be in the const section like MaxRSSPacketCount

PR #338, opened 2018-05-08
Modified style in merge function.

dkolistratova:
suggested by @Quasilyte

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -1165,7 +1165,7 @@ func send(parameters interface{}, flag *int32, coreID int) {
 	low.Send(srp.port, srp.queue, srp.in, flag, coreID)
 }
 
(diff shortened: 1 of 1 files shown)


Quasilyte:
I wonder what other team members think about this:
```diff
-func merge(from *low.Ring, to *low.Ring) {
+func merge(from, to *low.Ring) {
```

Type switch change looks like a clear win in terms of readability.

ifilippov:
This is not urgent. Let's firstly make new "merge" tests and then make sure that they pass with this.

dkolistratova:
@ifilippov , Current tests pass, it shouldn't change anything - just simplified syntax.

ifilippov:
If new merge test pass - then you can merge.

PR #344, opened 2018-05-09
Added variables to test framework and other improvements

gshimansky:
Variables allow arbitrary string replacement in tests command
lines. New section of test configuration now defines default variable
values, other variables with values may be added or overridden from
command line.

Added -test switch to allow selecting only one/several tests from
configuration file.

Improved pass/fail diagnostics.

The first experimental step of converting configuration files is done
on stability.json checksum tests.

### DIFF
--- a/test/framework/config.go
+++ b/test/framework/config.go
@@ -10,23 +10,11 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


ifilippov:
Checked, approved. I will rewrite other stability and performance tests for this semantic.

gshimansky:
PR deleted from master branch because it wasn't intended to be merged there

PR #345, opened 2018-05-10
Added variables to test framework and other improvements

gshimansky:
Variables allow arbitrary string replacement in tests command
lines. New section of test configuration now defines default variable
values, other variables with values may be added or overridden from
command line.

Added -test switch to allow selecting only one/several tests from
configuration file.

Improved pass/fail diagnostics.

The first experimental step of converting configuration files is done
on stability.json checksum tests.

### DIFF
--- a/test/framework/config.go
+++ b/test/framework/config.go
@@ -10,23 +10,11 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
 )
-
-type hostPort struct {
-	host string
-	port int
-}
-
-// HostsList is a slice of host:port pairs that are used to replace
-// values read from JSON config file.
-type HostsList []hostPort
-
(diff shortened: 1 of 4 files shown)


test/framework/config.go
```diff
@@ -145,9 +135,49 @@ func ReadConfig(fileName string, hl HostsList) (*TestsuiteConfig, error) {
...
 
+	// Merge variables from command line into default variable values
+	// from JSON configuration
+	for kkk, vvv := range vl {
+		config.Variables[kkk] = vvv
+	}
+
+	// Replace variables in test command lines
+	for ttt := range config.Tests {
+		for aaa := range config.Tests[ttt].Apps {
```
Quasilyte: Tripple letter variable names are unusual. `>.<`

gshimansky: Variables like this are easier to search/replace in text editor.

Issue #286, opened 2018-04-18
We need to clean Merge test


ifilippov:
Now we have three merges:

1) inside segment
2) outside segment
3) between outside and some branches from segment

Need to consider all these variants in testMerge

ifilippov:
It is done. We need sequential merges, however it is in the future.

PR #348, opened 2018-05-14
First version of new test merge

dkolistratova:
#286


test/stability/testMerge/merge_test.go
```diff
@@ -8,9 +8,30 @@ import (
...
+
+// TODO unfortunately this test can't be executed as go test right now due to https://github.com/intel-go/nff-go/issues/301
+// If you want to test it via go test you should firstly comment FreeMempools function at flow/flow.go
+// You will have multiple mempools however test will pass.
 func TestMerge(t *testing.T) {
-	if err := executeTest("", "", 0); err != nil {
-		t.Logf("fail: %+v\n", err)
-		t.Fail()
+	scenarios := []pipelineScenario{getGet, inCopy, copyCopy, segmSegm, segmGet, segmCopy, inSegm, manyFlows}
+	gens := []getScenario{fastGenerate}
```
mbulatova: Do we want to check generate case too?
If test become too long, can we run some tests with generate and some with fastGenerate?

test/stability/testMerge/testMerge.go
```diff
@@ -19,35 +18,50 @@ import (
...
+	receivePart
+	useGenerator1 = 0
+	useGenerator2 = 1
+)
+
 var (
-	totalPackets uint64 = 10000000
-	// Payload is 16 byte md5 hash sum of headers
+	totalPackets uint64 = 100000
+	// Payload is 16 byte
```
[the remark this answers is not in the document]
dkolistratova: Used in generator
```go
packet.InitEmptyIPv4UDPPacket(pkt, payloadSize)
```

test/stability/testMerge/merge_test.go
dkolistratova: We can't use generate with copy scenarios, cause speed is too large and test fails due to drops.
But with scenarios without copy it can be done. @ifilippov, do we really need it?

test/stability/testMerge/testMerge.go
```diff
@@ -19,35 +18,50 @@ import (
...
-// in the packet. Test also calculates sent/received ratios, number of broken
-// packets and prints it when a predefined number of packets is received.
+// This part of test generates packets sends on 2 ports and receives packets
+// on 1 port (merged by second part). Checks quantity and contents of packets according to scenario.
 //
 // Test with testScenario=2:
-// This part of test receives packets on 0 and 1 ports, merges flows
-// and send result flow to 0 port.
+// This part of test receives packets on 2 ports, merges flows
+// and send result flow to 1 port. There are several scenarios:
```
ifilippov: sends result to output port

test/stability/testMerge/testMerge.go
```diff
@@ -19,35 +18,50 @@ import (
...
+// 0: getGet - get packets, merge and send back
+// 1: inCopy - stop one of flows, copy the remaining, merge original flow with its copy and send back
+// 2: copyCopy - copy both flows, merge all 4 and send back
+// 3: segmSegm - add segment functions at both flows, merge and send
+// 4: segmGet - add segment function to one flow, merge and send
+// 5: segmCopy - stop one of flows, add segment function to the remaining, copy it , merge original flow with its copy and send back
+// 6: inSegm - stop one flow, split second, merge and send back
+// 7: manyFlows - split one flow, copy another and stop original, merge copy and two splitted, send
```
ifilippov: Please rename and modify:
0 -> 0
---
4 -> 1
3 -> 2 multipleSegm
6 -> 3 singleSegm
---
1 -> 4 getCopy (merge also with get, don't stop it)
2 -> 5 
7 -> 6 segmCopy
5 - It is part of sequential function and really is identical with 2 from the point of merge. Redundant here.

test/stability/testMerge/testMerge.go
```diff
@@ -89,151 +123,440 @@ func main() {
...
+		}
+		finalFlow = outputFlows[1]
+		*f = *outputFlows[0]
 	}
-	stabilityCommon.InitCommonState(configFile, target)
+	return finalFlow, err
+}
+
+func splitPackets(currentPacket *packet.Packet, context flow.UserContext) uint {
+	return 0
```
ifilippov: All packets will go to one flow. No merge checking will be here. If you don't want to add any split logic - change this to partition. No logic, but two flows. Profit!

test/stability/testMerge/testMerge.go
```diff
@@ -89,151 +123,440 @@ func main() {
...
+}
+
+func handlePackets(pkt *packet.Packet, context flow.UserContext) {
+}
+
+func filterPackets(pkt *packet.Packet, context flow.UserContext) bool {
+	return true
```
ifilippov: Also all packets to one part. Remove this. Handle and partition is quite enough.

test/stability/testMerge/testMerge.go
```diff
@@ -89,151 +123,440 @@ func main() {
...
+type segmentScenario uint
+
+const (
+	handle segmentScenario = iota
+	handleDrop
+	split
+)
+
+// returns nil if no new flow was created and new flow if was
+func setSegment(scenario segmentScenario, f *flow.Flow) (finalFlow *flow.Flow, err error) {
```
ifilippov: You change one function setSegment to one final function Set something. It is confusing. Please, remove setSegment function and set functions directly in their places. Also, if you post that you place "split" you should honestly test both flows. If you don't want to have split logic - place partition instead.

ifilippov:
Please, deal with current remarks plus rename commit to "renewal of merge test".
(also honestly you should change stability-VM.json file, however I will remove it, so don't do anything with it.)

test/stability/testMerge/testMerge.go
```diff
@@ -273,46 +587,115 @@ func generatePacketGroup2(pkt *packet.Packet, context flow.UserContext) {
...
 		atomic.StoreInt32(&passed, 0)
 	}
+	if atomic.AddUint64(&recvPackets, 1) >= totalPackets {
+		testDoneEvent.Signal()
+		return
+	}
+}
+
+// Count packets in copied flow
+func countCopiedPackets(pkt *packet.Packet, context flow.UserContext) {
```
ifilippov: We don't need this function here.
You set 0 to passed at different machine - it won't make any effect. Also you shouldn't check copy here. Any copy problems will be seen in Copy test.

ifilippov: Also you count copied packets at separate machine which won't produce any result. -> remove this function.

test/stability/testMerge/merge_test.go
ifilippov: We don't want to have generate here. It will be part of InputOutput tests. Fast generate is enough, you can remove slice and use only fast Generate everywhere.

test/stability/testMerge/testMerge.go
dkolistratova: I need handler, cause change packet's payload in it.
Counted packets are used in gotest.

dkolistratova:
Yes, I knew that stability-VM will be removed, that's why didn't do anything with it.

test/stability/testMerge/testMerge.go
```diff
@@ -71,13 +85,33 @@ var (
...
 func main() {
-	var testScenario uint
+	var pipelineType uint
+	var generatorType getScenario
+	var addSegmAfterMerge, useSimpleGenerate bool
+	flag.BoolVar(&useSimpleGenerate, "useSimpleGenerate", false, "true to use simple generator, false (default) to use fast generator")
+	flag.BoolVar(&addSegmAfterMerge, "addSegmAfterMerge", false, "true to insert handler after merge in pipeline, false(default) otherwise")
 	flag.UintVar(&testScenario, "testScenario", 0, "1 to use 1st part scenario, 2 snd, 0 to use one-machine test")
-	flag.Uint64Var(&passedLimit, "passedLimit", passedLimit, "received/sent minimum ratio to pass test")
+	flag.UintVar(&pipelineType, "pipelineType", 0, "merge between different ff types: 0 - getGet, 1 - inside Copy, 2- between Copy, 3 - between Segments, 4 - segm and get, 5 - segm and copy, 6 - inside Segm, 7 - between manyFlows")
```
ifilippov: ??? You have only 0-5 in stability.json

test/stability/testMerge/testMerge.go
ifilippov: please, rename this function to packetChaning or something like this. And use it everywhere, not only after copies if you want.
Remove copy counting stuff. It is not about merge. If you would like to count packets after copy you should write separate test which will be called only with gotest

ifilippov: It is a very bad practice to have tests which can PASS (after using test system), but test something ELSE when using gotest. What???

test/stability/testMerge/testMerge.go
```diff
@@ -89,151 +123,356 @@ func main() {
...
+		if inputSource == useGenerator1 {
+			finalFlow, err = flow.SetFastGenerator(generatePacketGroup, speed, &generatorParameters{isGroup1: true})
+		} else if inputSource == useGenerator2 {
+			finalFlow, err = flow.SetFastGenerator(generatePacketGroup, speed, &generatorParameters{isGroup1: false})
+		} else {
+			return nil, errors.New(" setGetter: unknown generator type")
+		}
+	default:
+		if inputSource == useGenerator1 {
+			finalFlow = flow.SetGenerator(generatePacketGroup, &generatorParameters{isGroup1: true})
```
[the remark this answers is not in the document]
dkolistratova: We decided to remove in gotest,, I thought the possibility to run manually should be left.

test/stability/testMerge/testMerge.go
```diff
@@ -60,7 +74,7 @@ var (
 
 	// T second timeout is used to let generator reach required speed
 	// During timeout packets are skipped and not counted
-	T = 10 * time.Second
+	T = 20 * time.Second
```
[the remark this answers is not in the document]
dkolistratova: All tests fail with low ratio if there is 10 seconds.

test/stability/testMerge/testMerge.go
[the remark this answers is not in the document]
ifilippov: And it will be our secret? No stability.json, no gotest and only three people in whole world know about this? Please, remove.

test/stability/testMerge/testMerge.go
```diff
@@ -89,151 +121,350 @@ func main() {
...
+	segmCopy
+)
+
+// set copier and handler to change packets in copied flow for future statistics
+func setCopyWithHandler(inFlow *flow.Flow) (*flow.Flow, error) {
+	outFlow, err := flow.SetCopier(inFlow)
+	if err != nil {
+		return nil, err
+	}
+	err = flow.SetHandler(outFlow, modifyCopiedPackets, nil)
```
ifilippov: When you modify packets this way you do the following:
1) Copy as copy
2) Handle as start of new segment
3) Merge handle's output
It is a different story when you merge segments. Here we would like to merge Copy. You use recvCopy1 and recvCopy2 only with "+recv1" which is strange. No PASS/FAIL conditions on them. You should honestly set Copy. Without any starting segment (without any handle after it). We don't need to know about copied flows. In the final proportion they should be "*2".

test/stability/testMerge/testMerge.go
```diff
@@ -89,151 +121,350 @@ func main() {
...
+	sent2 = atomic.LoadUint64(&sentPacketsGroup2)
+	shouldGet1 = sent1
+	shouldGet2 = sent2
+	switch scenario {
+	case singleSegm:
+		shouldGet2 = 0
+		lessPercent = 100
+		eps = 0
+	case getCopy:
+		shouldGetCopy1 = sent1
```
ifilippov: shouldGet1 *= 2

test/stability/testMerge/testMerge.go
```diff
@@ -89,151 +121,350 @@ func main() {
...
+	var sent1, sent2, shouldGet1, shouldGet2, shouldGetCopy1, shouldGetCopy2 uint64
+	sent1 = atomic.LoadUint64(&sentPacketsGroup1)
+	sent2 = atomic.LoadUint64(&sentPacketsGroup2)
+	shouldGet1 = sent1
+	shouldGet2 = sent2
+	switch scenario {
+	case singleSegm:
+		shouldGet2 = 0
+		lessPercent = 100
+		eps = 0
```
ifilippov: Really? Looks like it will be always fluky

test/stability/testMerge/testMerge.go
```diff
@@ -89,151 +121,350 @@ func main() {
...
+	println("got group 1", recv1, "packets")
+	println("got copied group 1", recvCopy1, "packets")
+	println("got group 2", recv2, "packets")
+	println("got copied group 2", recvCopy2, "packets")
+	println("=================================")
+
+	var ratio1, ratio2 uint64
+	if shouldGet1+shouldGetCopy1 != 0 {
+		ratio1 = (recv1 + recvCopy1) * 100 / (shouldGet1 + shouldGetCopy1)
+	} else {
```
ifilippov: You check this here: return failCheck("something went wrong..sent 0, but received not 0 in some of groups"). Please remove one of checkings.

test/stability/testMerge/testMerge.go
```diff
@@ -89,151 +121,350 @@ func main() {
...
+		if (recv1 + recvCopy1) != 0 {
+			return failCheck("lost all packets from group 1")
+		}
+	}
+
+	if shouldGet2+shouldGetCopy2 != 0 {
+		ratio2 = (recv2 + recvCopy2) * 100 / (shouldGet2 + shouldGetCopy2)
+	} else {
```
ifilippov: You check this here: return failCheck("something went wrong..sent 0, but received not 0 in some of groups"). Please remove one of checkings.

test/stability/testMerge/testMerge.go
dkolistratova: No, we send only one group of packets, thats why it will always be 100: 0 percent group ratios

dkolistratova: We check send\recv total ratio, this is a group ratio.

test/stability/testMerge/testMerge.go
```diff
@@ -19,35 +18,48 @@ import (
...
 	sentPacketsGroup2 uint64
-	recvPacketsGroup1 uint64
-	recvPacketsGroup2 uint64
-	recvPackets       uint64
-	brokenPackets     uint64
+
+	recvPacketsGroup1     uint64
+	recvPacketsGroup2     uint64
+	recvCopyPacketsGroup1 uint64
+	recvCopyPacketsGroup2 uint64
```
ifilippov: Remove

ifilippov:
Add copyCopy (merge all four flows) also in stability.json and approve.

Issue #349, opened 2018-05-14
When handler is inserted before copier packets start getting lost


dkolistratova:
Below is the code that works normally (ratio 100%), but if you uncomment the line with handler insertion, packets are dropped and ratio is 67%. 

```go
package main

import (
	"log"
	"sync"
	"sync/atomic"
	"time"

	"github.com/intel-go/nff-go/flow"
	"github.com/intel-go/nff-go/packet"
)

var (
	totalPackets uint64 = 100000
	// Payload is 16 byte
	payloadSize uint   = 16

	sentPackets   uint64
	recvPackets   uint64
	copiedPackets   uint64
	
	testDoneEvent *sync.Cond
	progStart     time.Time

	// T second timeout is used to let generator reach required speed
	// During timeout packets are skipped and not counted
	T = 20 * time.Second
)

func main() {
	var m sync.Mutex
	testDoneEvent = sync.NewCond(&m)
	config := flow.Config{
		DPDKArgs: []string{"--log-level=0"},
	}
	flow.CheckFatal(flow.SystemInit(&config))
	flow1, err := flow.SetFastGenerator(generatePacketGroup, 1000000, nil)
	flow.CheckFatal(err)

	// When line below is uncommented, large drops appear and ratio is 67% instead of 100.
	//flow.CheckFatal(flow.SetHandler(flow1, handlePackets, nil))
	
	flow1copy, err := flow.SetCopier(flow1)
	flow.CheckFatal(err)
	flow.CheckFatal(flow.SetHandler(flow1copy, countCopiedPackets, nil))
	outFlow, err := flow.SetMerger(flow1, flow1copy)
	flow.CheckFatal(err)
	flow.CheckFatal(flow.SetHandler(outFlow, checkPackets, nil))
	flow.CheckFatal(flow.SetStopper(outFlow))
	// Start pipeline
	go func() {
		err = flow.SystemStart()
	}()
	flow.CheckFatal(err)
	progStart = time.Now()

	// Wait for enough packets to arrive
	testDoneEvent.L.Lock()
	testDoneEvent.Wait()
	testDoneEvent.L.Unlock()
	println("sent: ", atomic.LoadUint64(&sentPackets))
	println("received: ", atomic.LoadUint64(&recvPackets))
	println("copied: ", atomic.LoadUint64(&copiedPackets))
[...]

PR #350, opened 2018-05-14
SetCopier fix

ifilippov:


ifilippov:
https://github.com/intel-go/nff-go/issues/349

ifilippov:
Summary of changes: if we aren't in segment before copy - it is the same. Just use currentRing of our flow. If we are inside segment we should finish it, make output ring and then create copy from this ring.

Issue #291, opened 2018-04-19
Latency performance test


ifilippov:
What is speed parameter? Can you please make this test more actual.

ifilippov:
Also you should change this test to generate some files (with modified perf_compare) which can be compared with previous ones.

mbulatova:
Current test is in my branch https://github.com/intel-go/nff-go/tree/mbulatova/latency
Added scope according to perf.json (perf_main, perf_light, perf_seq)
Takes 34m to complete

mbulatova:
After #353 is implemented, writing to files will not be necessary

PR #354, opened 2018-05-15
Update latency test, add to test framework

mbulatova:
- Calculate median, average and stddev latency
- Add writing statistics to file
- Modify test system to copy created output file from container to log dir
- Add timout to stabilize speed
- Use FastGenerator
- Remove latency-stub.go as unnecessary (perf_light and others should be
used as part2)
- Add more tests to latency.json
- Add noscheduler flag


mbulatova:
#291

ifilippov:
Approved, however you should connect with Gregory about test system changes.

mbulatova:
Updated

gshimansky:
That update is good, you got everything right, except that there is no report for latency tests. It is necessary to add a new section to HTML report in report.go and a report_test with a latency test type which you can use to debug the html output.
If you think that is too complex for you, you can commit the current change, and I will add the report part later myself.

mbulatova:
@gshimansky thanks for tips!

mbulatova:
@gshimansky is last commit ok?

Issue #356, opened 2018-05-16
VMs provisioned with Vagrant should have special name for system interface


gshimansky:
Currently VMs created by vagrant script have generic name for system management network interface (used for ssh) 'Wired connection 1'. It happens because after switching to consistent network names and rebooting the system cannot find eth0 interface and creates a generic name for every network connection.

The problem with generic names is that CI scripts remove them as garbage, and so remove system network interface as well.

PR #357, opened 2018-05-16
Fixed VM system network configuration to have special name

gshimansky:
This is necessary for testing scripts to work correctly because
generic name connections are wiped as garbage.
Fix for #356.


ifilippov:
Approved

Issue #352, opened 2018-05-14
Add support to Apache benchmark to test framework


gshimansky:
NAT performance tests use Apache benchmark. Test framework should be able to extract values from its output and add them to generated report.

PR #358, opened 2018-05-16
Added apache benchmark type of tests to report

gshimansky:
Fixes issue #352

### DIFF
--- a/test/framework/Makefile
+++ b/test/framework/Makefile
@@ -10,4 +10,7 @@ main: apptype_string.go
 %_string.go: types.go
 	go generate
 
+testing:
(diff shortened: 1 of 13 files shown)


ifilippov:
Dark, dark forest... OK. (If you add new test types, is it simple to merge with https://github.com/intel-go/nff-go/pull/354 ?)

gshimansky:
We discussed #354 latency tests with Maria and I convinced her that she needs to add a new test type in the same way as I added Apache Benchmark tests type. A new test type will gather information from tests by parsing their output and produce its own customized report.

Issue #237, opened 2018-03-15
Need to rewrite NAT


ifilippov:
Need to rewrite NAT using DirectSend for ARP requests, instead of confusing merges.

ifilippov:
Done.

PR #362, opened 2018-05-21
Improve ARP and ICMP handling in NAT example

ifilippov:


ifilippov:
https://github.com/intel-go/nff-go/issues/237

PR #365, opened 2018-05-22
Fix file permissions

mbulatova:


ifilippov:
If this is all, then approve.

PR #369, opened 2018-05-22
Added Mbits output

ifilippov:


gshimansky:
How does it affect speed?

ifilippov:
Will determine a little bit later

ifilippov:
Checked, performance impact is within margin of error.

ifilippov:
Performance is not effected with switched off by default:-). I didn't check with switched on.

PR #371, opened 2018-05-25
Fixed null ptr dereference when tests data is incomplete

gshimansky:


ifilippov:
Maybe it is OK if you think so.

PR #380, opened 2018-05-31
Added a note about connecting VMs via bridges

### LINKED ISSUE #379: Unreliable UDP tunnel connection between VMs cause tests to fail
Sometimes communication between VMs simply stops for an unknown reason.

gshimansky:
Attempt to fix #379


ifilippov:
If you think that it is OK - so OK.

Issue #387, opened 2018-06-08
Performance report comparison tool should be able to download reports from HTTP(S) server


gshimansky:
TeamCity stores reports on an artifactory, so report comparison tool should be able to retrieve reports from a web server.

PR #388, opened 2018-06-08
Fix for #387. Report comparison tool now supports remote files

gshimansky:


ifilippov:
Approved, however this comparison doesn't work at real files.

gshimansky:
Can you please attach real files to some new issue?

PR #396, opened 2018-06-14
Added core limitations to performance tests

ifilippov:
New scheduler model will allow more clones, so at some point framework
will start to use hyperthreading cores which is not expected behavior.


ifilippov:
Current scheduler exceeds 43 cores only in one test, so I don't expect any influence.

Issue #323, opened 2018-05-03
testCksum fails with flags "-udp -tcp -ipv4 -ipv6"


dkolistratova:
git bisect bad
8df5cdbeacd90f7d315d6ec7e12697eb0d90b61b is the first bad commit

reproducer:
```
 ./test/stability/testCksum/testCksum -udp -tcp -ipv4 -ipv6 -number 10000
```

fails with 
```
TEST FAILED
Bad data found in the packet: fdfe8aa4586acfce
```

ifilippov:
As Daria pointed these command line invocate random tests. Daria should provide exact line without random generations.

dkolistratova:
fixed

PR #397, opened 2018-06-18
Fixed wrong tcp data field setting which made cksum test fail. Small changes in test itself

dkolistratova:
#323 fix

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -79,6 +79,12 @@ const (
 	GTPMinLen  = 8
 )
 
+const (
+	TCPMinDataOffset = 0x50 // minimal tcp data offset
+	IPv4VersionIhl   = 0x45 // IPv4, IHL = 5 (min header len)
+	IPv6VtcFlow      = 0x60 // IPv6 version
+)
+
 // LogType - type of logging, used in flow package
 type LogType uint8
 
--- a/packet/packet.go
+++ b/packet/packet.go
@@ -490,7 +490,7 @@ func InitEmptyIPv4Packet(packet *Packet, plSize uint) bool {
 
 	// Next fields not required by pktgen to accept packet. But set anyway
 	packet.ParseL3()
-	packet.GetIPv4NoCheck().VersionIhl = 0x45 // Ipv4, IHL = 5 (min header len)
+	packet.GetIPv4NoCheck().VersionIhl = IPv4VersionIhl
 	packet.GetIPv4NoCheck().TotalLength = SwapBytesUint16(uint16(IPv4MinLen + plSize))
 	packet.GetIPv4NoCheck().NextProtoID = NoNextHeader
 	if hwtxchecksum {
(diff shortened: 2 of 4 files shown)


common/common.go
```diff
@@ -74,6 +74,7 @@ const (
 	IPv6Len    = 40
 	ICMPLen    = 8
 	TCPMinLen  = 20
+	TCPDataOff = 0x50
```
mbulatova: It is actually TCPMinDataOffset

packet/packet.go
```diff
@@ -554,8 +553,8 @@ func InitEmptyIPv4TCPPacket(packet *Packet, plSize uint) bool {
 	packet.GetIPv4NoCheck().TotalLength = SwapBytesUint16(uint16(IPv4MinLen + TCPMinLen + plSize))
 
 	packet.ParseL4ForIPv4()
-	packet.GetTCPNoCheck().DataOff = packet.GetTCPNoCheck().DataOff | 0x50 // TODO check
-
+	packet.GetTCPNoCheck().DataOff = TCPDataOff
+	packet.Data = unsafe.Pointer(uintptr(packet.L4) + uintptr(((*TCPHdr)(packet.L4)).DataOff&0xf0)>>2)
```
gshimansky: Why not use packet.GetTCPNoCheck().DataOff here?

packet/packet.go
```diff
@@ -626,16 +625,15 @@ func InitEmptyIPv6TCPPacket(packet *Packet, plSize uint) bool {
...
 	packet.ParseL3()
 	packet.GetIPv6NoCheck().Proto = TCPNumber
 	packet.GetIPv6NoCheck().PayloadLen = SwapBytesUint16(uint16(TCPMinLen + plSize))
 	packet.GetIPv6NoCheck().VtcFlow = SwapBytesUint32(0x60 << 24) // IP version
 
 	packet.ParseL4ForIPv6()
-	packet.GetTCPNoCheck().DataOff = packet.GetTCPNoCheck().DataOff | 0x50
-
+	packet.GetTCPNoCheck().DataOff = TCPDataOff
+	packet.Data = unsafe.Pointer(uintptr(packet.L4) + uintptr(((*TCPHdr)(packet.L4)).DataOff&0xf0)>>2)
```
gshimansky: Why not use packet.GetTCPNoCheck().DataOff here?

common/common.go
gshimansky: I agree.

ifilippov: If you start to add new constants not connected with length, can you
1) start const section
2) add constants
0x45 // Ipv4, IHL = 5 (min header len) - 610 line
SwapBytesUint32(0x60 << 24) // IP version - 632 line
and others.

test/stability/testCksum/testCksum.go
```diff
@@ -47,7 +47,7 @@ const (
 )
 
 var (
-	totalPackets uint64 = 10
+	totalPackets uint64 = 32
```
ifilippov: Was this tested both at physical machines and virtual machines?

dkolistratova: yes. 32 is more logical cause its the burst size.

dkolistratova:
failed due to dpdk site unavailable. Checked locally, works.

PR #401, opened 2018-06-19
Refactored packet initialization to remove duplicate code.

dkolistratova:
Removed a bunch of duplicates (was noticed changing constants in pr #397 ) by the way, there was also a bug  packet.GetIPv6NoCheck().PayloadLen = SwapBytesUint16(uint16(UDPLen + plSize)) for icmp packet.

### DIFF
--- a/packet/packet.go
+++ b/packet/packet.go
@@ -473,6 +473,18 @@ func InitEmptyPacket(packet *Packet, plSize uint) bool {
 	return true
 }
 
+func fillIPv4Default(packet *Packet, plLen uint16, nextProto uint8) {
+	packet.GetIPv4NoCheck().VersionIhl = IPv4VersionIhl
+	packet.GetIPv4NoCheck().TotalLength = SwapBytesUint16(plLen)
+	packet.GetIPv4NoCheck().NextProtoID = nextProto
+}
+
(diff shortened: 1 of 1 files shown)


ifilippov:
You should remove ethernet and ParseL3 from your functions
You should pass packet.GetIPv4NoCheck() to you function instead of packet

dkolistratova:
@ifilippov 

then packet.Ether.EtherType = SwapBytesUint16(IPV6Number)
	packet.ParseL3()
lines will be duplicated everyrhere. 

Maybe call function "fillCommonDataForIPv6" and leave as is?

ifilippov:
Why are you care about ParseL3 and don't care about ParseL4 which is also duplicated? I don't like an idea to move Parses anywhere. It is OK to duplicate them. Also I don't like an idea to move ethernet initialization anywhere because it is confusing.

packet/packet.go
```diff
@@ -473,6 +473,18 @@ func InitEmptyPacket(packet *Packet, plSize uint) bool {
 	return true
 }
 
+func fillIPv4(packet *Packet, plLen uint16, nextProto uint8) {
```
ifilippov: Word default somewhere. fillIpv4Default for example.

ifilippov:
Approved, you can merge.

Issue #413, opened 2018-06-28
NFF-GO hangs when speed is large and generator is heavy


dkolistratova:
Reproducer:
```go
package main

import (
	"fmt"
	"time"

	"github.com/intel-go/nff-go/flow"
	"github.com/intel-go/nff-go/packet"
)

func main() {
	config := flow.Config{}
	flow.CheckFatal(flow.SystemInit(&config))
	outFlow, err := flow.SetFastGenerator(generate, 1000000000, nil)
	flow.CheckFatal(err)
	flow.CheckFatal(flow.SetSender(outFlow, uint16(0)))
	flow.CheckFatal(flow.SystemStart())
}


func generate(pkt *packet.Packet, context flow.UserContext) {
	if pkt == nil {
		panic("Failed to create new packet")
	}
	if !packet.InitEmptyPacket(pkt, 16) {
		panic(fmt.Sprintf("InitEmptyPacket failed"))
	}
	time.Sleep(100000)
}
```
run:
```
./tmp
```

Output:
```
DEBUG: ---------------
DEBUG: Start new clone for fast generator instance 0 at 43 core
DEBUG: Start new clone for fast generator instance 0 at 44 core
DEBUG: System is using 45 cores now. 43 cores are left available.
DEBUG: Current speed of fast generator is 194240 PKT/S, target speed is 1000000000 PKT/S
DEBUG: Mempool usage fast generate 1408 from 8191
DEBUG: Mempool usage slow operations 0 from 8191
DEBUG: ---------------
DEBUG: Start new clone for fast generator instance 0 at 45 core
DEBUG: System is using 46 cores now. 42 cores are left available.
DEBUG: Current speed of fast generator is 197248 PKT/S, target speed is 1000000000 PKT/S
DEBUG: Mempool usage fast generate 1184 from 8191
DEBUG: Mempool usage slow operations 0 from 8191
DEBUG: ---------------
DEBUG: Start new clone for fast generator instance 0 at 46 core
DEBUG: Start new clone for fast generator instance 0 at 47 core
DEBUG: System is using 48 cores now. 40 cores are left available.
DEBUG: Current speed of fast generator is 200256 PKT/S, target speed is 1000000000 PKT/S
DEBUG: Mempool usage fast generate 640 from 8191
DEBUG: Mempool usage slow operations 0 from 8191
DEBUG: ---------------
DEBUG: Start new clone for fast generator instance 0 at 48 core
```
And then nothing. 
lscpu:
[...]

PR #418, opened 2018-07-10
Fix number of report channels

ifilippov:
For https://github.com/intel-go/nff-go/issues/413


flow/scheduler.go
```diff
@@ -583,6 +583,12 @@ func (ff *flowFunction) updateCurrentSpeed() {
 			ffi.currentSpeed.Packets += temp.Packets
 			ffi.currentSpeed.Bytes += temp.Bytes
 		}
+		t = len(ffi.report)
```
aregm: Add comment here

PR #424, opened 2018-07-17
NGIC DP Implementation

vinodborole:
NGIC-DP
   1] NGIC DP implementation w/S11
        --CP Simulator populating the UE context with UL and DL Maps
   2] ADC, SDF, PCC static rules
        --Static rules are parsed and created ACLs 
   3] KNI with static/dynamic ARP support
        --Used netlink library for route and arp discovery bringing in the linux arp and route table in the fast   path

Flow Diagram:
> ```
>                     +------------------+
>       Uplink (UL)   |                  |
>       ------------> +                  +---------->
>                     |                  |
>         GTPU        | S1U          SGI |      UDP
>                     |                  |
>    Downlink(DL)     |                  |
>       <------------+                   +<----------
>                     +------------------+
> ```


ifilippov:
Hi All,

Overall it looks good, great work! However I have two types of questions - structure and code. Let's begin with structure:

Out idea is to have your example as example, not as part of library that will be imported by users. You implemented lots of great things and we will be very appreciate if you:
1) move darp/dynamicarp.go and sarp/staticarp.go code into our packet/arp.go file. It doesn't have any tables and netlink now - so it will be very useful. (Parts of these files which are only parts of NGIC example should remain in NGIC directory)
2) make new highest level package rules and place adcrule.go, constants.go, pccrule.go and sdfrule.go files there. We will then move acl.go from current packet package to this new package to have all rules at one place.

Also I suggest
1) remove global package functionality because packages should be independent (maybe this will require some structural changes)
2) remove obvious parts from README. However add all information about kni.sh scripts and other unusual things. I really can't understand some settings now.
3) maybe it will be better to move kni scripts inside run.sh under options?
4) add Dockerfile as in other examples

Can you please point me if you are against any of these suggestions or if they are impossible or not logical?

What about code moments - we will make one more round of review for this. I have some questions about comments, CheckFatal, etc.

Thank you for your contribution!

aregm:
@vinodborole - also is there a need to have 2 commits?

vinodborole:
@aregm The commit is just one, however there is one merge commit which can be ignored

vinodborole:
@amitinfo2k Can you check on the suggestions given by @ifilippov ?

rules/adcrule.go
```diff
@@ -0,0 +1,151 @@
+//Package rules ...
+// Copyright (c) 2017 Intel Corporation.
```
ifilippov: (c) 2018 and remove ... everywhere

rules/adcrule.go
```diff
@@ -0,0 +1,151 @@
...
+
+			_, ipnet, err := net.ParseCIDR(adcFilter.IP + "/" + strconv.Itoa(adcFilter.Prefix))
+			flow.CheckFatal(err)
+
+			adcFilter.IPPrefixNW = *ipnet //ip-net will ease the lookup
+		}
+
+		if section.HasKey(LblDomain) {
+			adcFilter.Domain = section.Key(LblDomain).String()
+
```
ifilippov: Please, use "gofmt -s -w" for all .go files

common/common.go
```diff
@@ -93,6 +95,8 @@ const (
 	No LogType = 1 << iota
 	// Initialization - output during system initialization
 	Initialization
+	//
+	Info
```
ifilippov: We use "Debug" for the same purposes

common/common.go
```diff
@@ -305,10 +317,16 @@ func LogTitle(logType LogType, v ...interface{}) {
 
 // SetLogType internal, used in flow package
 func SetLogType(logType LogType) {
-	log.SetFlags(0)
+//	log.SetFlags(0)
```
ifilippov: Why?

common/common.go
```diff
@@ -412,3 +430,66 @@ func dropInvalidCPUs(nums []int, maxcpu int) []int {
...
+
+//convert int ip to  net.IP
+func Int2ip(nn uint32) net.IP {
+	ip := make(net.IP, 4)
+	binary.BigEndian.PutUint32(ip, nn)
+	return ip
+}
+
+// byte arr to ipv4 int
+func ByteAryToIPv4(ipB [4]byte) uint32 {
```
ifilippov: We have this function in packet/packet.go

common/common.go
```diff
@@ -412,3 +430,66 @@ func dropInvalidCPUs(nums []int, maxcpu int) []int {
...
+	return uint32(ipB[3])<<24 | uint32(ipB[2])<<16 | uint32(ipB[1])<<8 | uint32(ipB[0])
+}
+//util method to print MAC
+func PrintMAC(prompt string, mac [EtherAddrLen]uint8) {
+	str_map := fmt.Sprintf("%s: %02x:%02x:%02x:%02x:%02x:%02x", prompt, mac[0], mac[1], mac[2], mac[3], mac[4], mac[5])
```
ifilippov: Use MACToString from packet/packet.go

common/common.go
```diff
@@ -412,3 +430,66 @@ func dropInvalidCPUs(nums []int, maxcpu int) []int {
 	}
 	return nums[:i]
 }
+// parse ip address string to int ip
+func StringToIPv4(ipaddr string) uint32 {
```
ifilippov: Move this function to packet/packet.go

rules/rule.go
```diff
@@ -0,0 +1,58 @@
...
+	sdfFilters []SDFFilter
+)
+
+var (
+	pccSDFMap map[int][]PCCFilter
+	pccADCMap map[int][]PCCFilter
+)
+
+//Init ... initial rules and prepare and install acl
+func Init() {
```
ifilippov: This should be somewhere in example, not as part of public library file? (Am I wrong?)

rules/pccrule.go
```diff
@@ -0,0 +1,145 @@
...
+
+	for i := 1; i <= noOfFilters; i++ {
+		strVal := strconv.Itoa(i)
+
+		section := cfg.Section(LblPccFilter + "_" + strVal)
+		pccFilter := PCCFilter{PCCFilterIdx: i}
+
+		if section.HasKey(LblUlAmbrMtrProfileIdx) {
+			pccFilter.UlMbrMtrProfileIdx, err = section.Key(LblUlAmbrMtrProfileIdx).Int()
+			flow.CheckFatal(err)
```
ifilippov: You can CheckFatal only one time after all conditions

rules/constants.go
```diff
@@ -0,0 +1,78 @@
+//Package rules ...
+// Copyright (c) 2017 Intel Corporation.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+package rules
+
+// constants for rule file paths
+const (
```
ifilippov: You can't have constants for file paths in file that can be imported by user, can you please move it to example

ifilippov:
Can you please
deal with above points
use "gofmt -s -w" for all files
use golint for all files
merge to one commit (or one + merge)
Add new example directory to examples/Makefile
Add .gitignore file for example binaries

common/common.go
amitinfo2k: This was done for having separate logs for ngic dp to go in a log file, as previous comments suggested to reuse the logger so, for now i am using the different log level for it. we will find a better way to do it

common/common.go
amitinfo2k: I will recheck this was done for the reuse of the existing logger and separate logs from stats and ngic.

common/common.go
amitinfo2k: ok

common/common.go
amitinfo2k: ok

common/common.go
amitinfo2k: ok

rules/rule.go
amitinfo2k: ok

rules/constants.go
amitinfo2k: these constants are specific to rules pkg only. can be moved to rules.go file instead of example.

rules/adcrule.go
amitinfo2k: i used gofmt -w option may be i missed it let me do it again. I think it's better to integrate gofmt and golint in makefile. so, build itself will catch it.

rules/adcrule.go
amitinfo2k: ok

rules/pccrule.go
amitinfo2k: hmm let me rework it

rules/adcrule.go
ifilippov: Anyway, remove empty lines before closing brackets

rules/rule.go
amitinfo2k: will need to review it

common/common.go
ifilippov: It is confusing - it is library and NGIC is an example. Why one example shouldn't use DEBUG logs to be separated?

common/common.go
ifilippov: I don't understand why do you want to separate logs. Also you can't change library files to satisfy one example.

common/common.go
ifilippov: Also please move PrintMAC to packet.go

common/common.go
amitinfo2k: We don't wants the stats in the logs as it's too clumsy so. had separate logger earlier. this was done as per the previous PR comment  "reuse existing logger and modify as needed ".

amitinfo2k: also, other examples are continue to use the logging as was earlier

rules/constants.go
amitinfo2k: ok removed the constants.go Instead of separated example dir i have moved them to appropriate files and path can be configured in ngicdp.go

rules/rule.go
amitinfo2k: taken care

PR #427, opened 2018-07-19
Add timers

ifilippov:


examples/timer.go
```diff
@@ -0,0 +1,49 @@
...
+		check = t.AddVariant(nil)
+	}
+	*check = true
+}
+
+func react(context flow.UserContext) {
+	fmt.Println("2 seconds after last packet was arrived")
+
+	answerPacket, _ := packet.NewPacket()
+	packet.InitEmptyIPv4Packet(answerPacket, 100)
```
aregm: Why 100? Make it minimal 64 or typical, something like 512

examples/timer.go
```diff
@@ -0,0 +1,49 @@
...
+
+func handler(currentPacket *packet.Packet, context flow.UserContext) {
+	if check == nil {
+		check = t.AddVariant(nil)
+	}
+	*check = true
+}
+
+func react(context flow.UserContext) {
+	fmt.Println("2 seconds after last packet was arrived")
```
aregm: Why 2 ? Can you make it parameter?

PR #433, opened 2018-07-24
Netlink example: can cache kernel routes into lpm, update it and lookup routes

dkolistratova:


examples/netlink.go
```diff
@@ -0,0 +1,159 @@
...
+var lpm *packet.LPM
+
+// lpm struct binds ip with number and below
+// is a struct to get ip by binded value
+var portToIP []uint32
+
+// length of portToIp
+var lenP2IP uint32
+
+const (
```
aregm: I think we should be consistent with the examples: make them external parameters with default values, as everywhere else.

PR #445, opened 2018-08-07
Added linking with dpdk ena lib for aws

dkolistratova:


ifilippov:
Check normal machines

Issue #440, opened 2018-08-02
All errors encountered in low.InitDPDK are fatal


gshimansky:
Errors in initialization should not be fatal and lead to application termination. They should be returned to the calling application.

dkolistratova:
done

PR #452, opened 2018-08-13
Make init return error instead of fail

dkolistratova:
#440

### DIFF
--- a/low/low.go
+++ b/low/low.go
@@ -487,11 +487,17 @@ func InitDPDKArguments(args []string) (C.int, **C.char) {
 }
 
 // InitDPDK initializes the Environment Abstraction Layer (EAL) in DPDK.
-func InitDPDK(argc C.int, argv **C.char, burstSize uint, mbufNumber uint, mbufCacheSize uint, needKNI int) {
-	C.eal_init(argc, argv, C.uint32_t(burstSize), C.int32_t(needKNI))
-
+func InitDPDK(argc C.int, argv **C.char, burstSize uint, mbufNumber uint, mbufCacheSize uint, needKNI int) error {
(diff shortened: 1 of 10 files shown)


low/low.go
```diff
@@ -635,9 +641,13 @@ func ReportMempoolsState() {
 }
 
 // CreateKni creates a KNI device
-func CreateKni(portId uint16, core uint, name string) {
+func CreateKni(portId uint16, core uint, name string) error {
 	mempool := (*C.struct_rte_mempool)(CreateMempool("KNI"))
-	C.create_kni(C.uint16_t(portId), C.uint32_t(core), C.CString(name), mempool)
+	ret := C.create_kni(C.uint16_t(portId), C.uint32_t(core), C.CString(name), mempool)
```
ifilippov: either use ret variable in returned error or remove it

low/low_test.go
```diff
@@ -9,10 +9,12 @@ import "testing"
 import "math/rand"
 import "time"
 
-func init() {
+func TestInit(t *testing.T) {
```
ifilippov: Why?

dkolistratova: to use t.FatalF

ifilippov: Bad practice. If it wasn't initialized test will also fail.
In you case what if user use go test -test=TestReorder ?

ifilippov: use log.Fatal if you want. Or nothing.

ifilippov:
Add checking to kni.go example. Than change low test and you can commit.

dkolistratova:
kni example already has this check cause CreateKniDevice returned error before

PR #454, opened 2018-08-14
More accurate synchronization between scheduler and handlers

ifilippov:
Changes:
1) time.Tick was changed to time.NewTicker due to garbage collector. From documentation: "While Tick is useful for clients that have no need to shut down the Ticker, be aware that without a way to shut it down the underlying Ticker cannot be recovered by the garbage collector; it "leaks"."
2) Add function synchronization was added. Now scheduler will wait until new clone really starts.
3) The main one: we change our synchronization policy:
a) We can't use strict synchronization between scheduler and handler due to performance reasons. So we have the following one:
b) If scheduler makes any changes with any clone of an instance or the instance in general we stopped tickers for all clones and start them from the beginning. (use stopper channel for this)
c) Otherwise handlers sends their statistic each sched_time period and scheduler reads it also any sched_time period. However scheduler makes some other work like starting and removing clones, debug and others. So after some time it will fall behind. We are OK with this. We will synchronize them any time from point (b). Other times little desync will not influence scheduler heuristics.

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -1010,7 +1010,6 @@ func segmentProcess(parameters interface{}, inIndex []int32, stopper [2]chan int
 		countOfPackets[index] = 0
 	}
 	var currentSpeed reportPair
-	tick := time.Tick(time.Duration(schedTime) * time.Millisecond)
 	var pause int
 	firstFunc := lp.firstFunc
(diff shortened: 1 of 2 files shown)


ifilippov:
Performance was not changed.

Issue #434, opened 2018-07-25
Current heuristic for stopping instance of segmentCopy function is wrong


ifilippov:

ifilippov:
We need to remove vector of speeds and replace it with number of "waiting time". Which is sum of zero packet taking attempts multiplied to constant time of one attempt.

ifilippov:
Done. However

"As Gregory pointed average time measure functionality can be replaced by easier and more accurate time measurements of every loop iterations. Tests shows that this is true with no performance degradation. However due to some problems with calculating real time it can't be done on the fly. So I will merge this and then change to accurate time measurements."

ifilippov:
After some performance investigations it appears that we need 3 "time.Now" for time measurements in critical path. Each time.Now is approximately 70nn so if user function is empty it is +9%, if user function has ten iterations loop with sums per packet it is +4%. This change is legal, however we postpone it for now. I close this.

PR #458, opened 2018-08-15
Core scheduler changes

ifilippov:
Before
Reported from handler:
- current speed
Was used in heuristics:
- clone removing:    as vector of speeds for all number of clones, removed clone is current is less than saved
                     really unstable heuristic
- clone adding:      as speed for +1 clone. Do not add clone if saved increased speed is less than current
- instance removing: usage was wrong
- instance adding:   no usage

After
Reported from handler:
- current speed
Is used in heuristics:
- clone removing:    as "decreased speed". If current speed right after addition is less than saved - remove
- clone adding:      as "increased speed" in above - no changes
+
Reported from handler:
- current number of zero attempts to get packets from input ring
Is used in heuristics:
- clone removing:    if "summ of all zero attempts" * "time per attempt (previously calculated)" > "schedtime" - remove
- instance removing: 1) get number of zero attempts from the ring with less number of them at each instance
                     2) convert this number to time by multiplying each instance number to corresponding time per attempt
                     3) get two instances with the biggest times
                     4) if summ of thier times is bigger than schedTime - combine them at one core
Logic for removing is the following. Each core has useful time and vacant time.
Vacant time is determined as time for zero attempts. If summ of vacant times of two cores is more than summ of their
usefull times then one core will be able to handle tasks of both cores. They can be combined.
We need to measure "time per attempT" before using it - it is quite flexible.

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -1009,7 +1009,7 @@ func segmentProcess(parameters interface{}, inIndex []int32, stopper [2]chan int
 		OutputMbufs[index] = make([]uintptr, burstSize)
 		countOfPackets[index] = 0
 	}
-	var currentSpeed reportPair
+	var currentState reportPair
 	var pause int
 	firstFunc := lp.firstFunc
 	// For scalar part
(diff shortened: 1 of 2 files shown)


ifilippov:
No performance changes

ifilippov:
Solved issue: https://github.com/intel-go/nff-go/issues/434

ifilippov:
As Gregory pointed average time measure functionality can be replaced by easier and more accurate time measurements of every loop iterations. Tests shows that this is true with no performance degradation. However due to some problems with calculating real time it can't be done on the fly. So I will merge this and then change to accurate time measurements.

PR #465, opened 2018-08-22
Added KNI, port forwarding and full ARP support to NAT

gshimansky:
- Cosmetic code refactoring into smaller separate functions
- Implemented KNI and port forwarding config
- Added creation of KNI devices and flows (no traffic forwarded yet)
- Implemented port forwarding (including KNI). It works to ssh from public interface.
- Direct all ARP and ICMP traffic to KNI if it is present on interface
- Fixed handling of ICMP and directing all ICMP traffic to KNI if it is present
- Fixed directing traffic to KNI in private subnet
- Implemented ARP requests when target destination MAC is unknown
- Fixed getting ARP response when KNI is enabled on interface
- Fixed directing traffic to KNI when it originates from NAT host

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -1614,8 +1614,8 @@ func (timer *Timer) Stop() {
 func CheckFatal(err error) {
 	if err != nil {
 		if nfErr := common.GetNFError(err); nfErr != nil {
-			common.LogFatalf(common.Debug, "failed with message and code: %+v\n", nfErr)
+			common.LogFatalf(common.No, "failed with message and code: %+v\n", nfErr)
 		}
-		common.LogFatalf(common.Debug, "failed with message: %s\n", err.Error())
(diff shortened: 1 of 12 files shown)


flow/flow.go
```diff
@@ -1614,8 +1614,8 @@ func (timer *Timer) Stop() {
 func CheckFatal(err error) {
 	if err != nil {
 		if nfErr := common.GetNFError(err); nfErr != nil {
-			common.LogFatalf(common.Debug, "failed with message and code: %+v\n", nfErr)
+			common.LogFatalf(common.No, "failed with message and code: %+v\n", nfErr)
```
ifilippov: Why?

gshimansky: This is to make sure that fatal error is printed even when debug log level is disabled.

PR #466, opened 2018-08-23
Fixed dpi with hyperscan tag

dkolistratova:
Temporary solution according to hyperscan fix. Will we use it or leave dpi example commented?

### DIFF
--- a/examples/dpi/Makefile
+++ b/examples/dpi/Makefile
@@ -5,4 +5,6 @@
 PATH_TO_MK = ../../mk
 SUBDIRS = main pattern
 
+export GO_COMPILE_FLAGS += -tags hyperscan_v4
+
 include $(PATH_TO_MK)/intermediate.mk
--- a/mk/leaf.mk
+++ b/mk/leaf.mk
@@ -12,7 +12,7 @@ include $(PATH_TO_MK)/include.mk
 
 ifdef NFF_GO_DEBUG
 # Flags to build Go files without optimizations
-export GO_COMPILE_FLAGS=-gcflags '-N -l'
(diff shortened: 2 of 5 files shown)


mk/leaf.mk
```diff
@@ -16,7 +16,7 @@ export GO_COMPILE_FLAGS=-gcflags '-N -l'
 endif
 
 $(EXECUTABLES) : % : %.go
-	go build $(GO_COMPILE_FLAGS) $< $(COMMON_FILES)
+	go build -tags hyperscan_v4 $(GO_COMPILE_FLAGS) $< $(COMMON_FILES)
```
gshimansky: I don't think it is a good approach to add `-tags hyperscan_v4` to a general build rule used to build all Go sources in the framework. Instead I suggest that you define `GO_COMPILE_FLAGS=-tags hyperscan_v4` in the beginning of `examples/dpi/pattern/Makefile` before including `leaf.mk`. This should add necessary switch only to compilation of DPI example.

dkolistratova: really good catch!

examples/dpi/Makefile
```diff
@@ -5,4 +5,6 @@
 PATH_TO_MK = ../../mk
 SUBDIRS = main pattern
 
+export GO_COMPILE_FLAGS += -tags hyperscan_v4
```
gshimansky: You should also change leaf.mk:15 to add `gcflags` to `GO_COMPILE_FLAGS` instead of assigning them. Otherwise with NFF_GO_DEBUG this will not work.

dkolistratova: Done

PR #475, opened 2018-09-10
Implemented DHCP client functionality to acquire IPv4 address

gshimansky:


scripts/get-depends.sh
```diff
@@ -8,6 +8,7 @@ go get -v golang.org/x/net/proxy
 go get -v github.com/docker/go-connections
 go get -v golang.org/x/tools/cmd/stringer
 go get -v github.com/vishvananda/netlink
+go get -v github.com/google/gopacket
```
ifilippov: If we add gopacket to dependence - remove it download from gopacket example.

PR #484, opened 2018-10-02
Removed NAT example because it is moved to nff-go-nat repository

gshimansky:


ifilippov:
Perf.json ? Stability ?

gshimansky:
> Perf.json ? Stability ?

Ok, removed too.

PR #491, opened 2018-10-23
Added IPv6 docker setup

gshimansky:


vagrant/scripts.sh
```diff
@@ -131,14 +131,28 @@ setupdocker ()
...
+[Service]
+Environment="HTTP_PROXY=${http_proxy}"
+EOF'
+    fi
+
     sudo mkdir /etc/docker
     sudo sh -c 'cat > /etc/docker/daemon.json <<EOF
 {
-    "hosts": ["unix:///var/run/docker.sock", "tcp://0.0.0.0:2375"]
+    "hosts": ["unix:///var/run/docker.sock", "tcp://0.0.0.0:2375"],
```
nplanel: untested but the ipv6 listen address should be "tcp://[::]:2375"

gshimansky: Thank you for suggestion but apparently it is redundant. Docker already listens to tcp6 when ipv6 is enabled
```
sudo netstat -tap | grep docker
tcp6       0      0 [::]:docker             [::]:*                  LISTEN      23833/dockerd
```
and if I add `"tcp://[::]:2375"` it fails to start
```
Failed to load listeners: listen tcp 0.0.0.0:2375: bind: address already in use
```

PR #501, opened 2018-11-29
Switched target development platform to Ubuntu 18.04

gshimansky:


gshimansky:
There are by far more common parts than differences in provisioning for Fedora and Ubuntu, so it is better to keep common stuff in one place.

aregm:
That is fine - just don't remove Fedora stuff at all. Have different file for that.

gshimansky:
Fedora stuff is not removed, it is still present in section `$provision_fedora`. Just currently it is not called because it is marked as `run: "never"`. To switch back to Fedora it is necessary to edit the file suppress Ubuntu and allow fedora specific provisioning to run.

PR #506, opened 2018-12-12
Add GetPortByName and GetNameByPort

blitzrk:
These are simple wrappers to `rte_eth_dev_get_port_by_name` and `rte_eth_dev_get_name_by_port` that I found useful. Not sure the best way to write tests for these funcs, so I figured I'd start this discussion to see if you even want to add these features.

I've found `GetPortByName` especially useful, because when a DPDK driver is bound, the easiest way to get a device's MAC address is to get its DPDK port by PCI address and then pass that to `low.GetPortMACAddress`.

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -974,6 +974,25 @@ func GetPortMACAddress(port uint16) [common.EtherAddrLen]uint8 {
 	return low.GetPortMACAddress(port)
 }
 
+// GetPortByName gets the port id from device name. The device name should be
+// specified as below:
+//
+// - PCIe address (Domain:Bus:Device.Function), for example- 0000:2:00.0
+// - SoC device name, for example- fsl-gmac0
+// - vdev dpdk name, for example- net_[pcap0|null0|tap0]
+func GetPortByName(name string) (uint16, error) {
+	return low.GetPortByName(name)
+}
+
+// GetNameByPort gets the device name from port id. The device name is specified as below:
+//
+// - PCIe address (Domain:Bus:Device.Function), for example- 0000:02:00.0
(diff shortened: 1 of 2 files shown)


ifilippov:
Hi blitzrk,

Can you please also add these functions to flow package. We encourage users not to import low package directly.

blitzrk:
I moved from low to flow and tested that it compiled with this command:

```
CGO_CFLAGS="-I$RTE_SDK/$RTE_TARGET/include -O3 -std=gnu11 -m64 -march=native -pthread -mno-fsgsbase -mno-f16c -include rte_config.h -Wno-deprecated-declarations" CGO_LDFLAGS_ALLOW=".*" CGO_LDFLAGS="-L$RTE_SDK/$RTE_TARGET/lib -Wl,--no-as-needed -Wl,-export-dynamic" GO111MODULE=on go build ./flow
```

This is the first time that `flow` imports the `C` pseudo package. That syntax wasn't intended to be contained to the `low` package?

ifilippov:
No. NFF-GO architecture is the following: low package imports "C" with all cgo dependencies. Than low functions are called from flow package by exported functions. Look for example "FreeMempools" function which is called from SystemStop or "ReceiveKNI" that is called from "recvKNI". You should simply add wrappers to flow package. All logic including cgo calls should remain in low package.

I'm sorry, I was not clear in first explanation.

ifilippov:
You should remain your first commit - it was good. And also add functions to flow.go that will call your new functions from low.go

blitzrk:
Added the wrappers and rebased to the latest develop branch.

blitzrk:
@aregm Updated my commit message to the requested format.

PR #514, opened 2018-12-15
Added special types for IPv4, IPv6 and MAC addresses

gshimansky:


ifilippov:
I propose to have this stuff in packet/addresses.go inside package packet. Because the idea is to do all work with packets without importing additional packages. It is strange, that parts of main structures like IPv4Hdr is designed in different package.

common/common.go
```diff
@@ -90,6 +92,85 @@ const (
...
+	return IPv4Address(a[3])<<24 | IPv4Address(a[2])<<16 | IPv4Address(a[1])<<8 | IPv4Address(a[0])
+}
+
+// SliceToIPv4 converts four element slice to IPv4Address representation
+func SliceToIPv4(s []byte) IPv4Address {
+	return IPv4Address(s[3])<<24 | IPv4Address(s[2])<<16 | IPv4Address(s[1])<<8 | IPv4Address(s[0])
+}
+
+// IPv4ToBytes converts four element address to IPv4Address representation
+func IPv4ToBytes(v IPv4Address) [IPv4AddrLen]byte {
```
ifilippov: IPv4ToArray

common/common.go
```diff
@@ -90,6 +92,85 @@ const (
...
+	return [IPv4AddrLen]uint8{byte(v), byte(v >> 8), byte(v >> 16), byte(v >> 24)}
+}
+
+func (addr IPv4Address) String() string {
+	return fmt.Sprintf("%d.%d.%d.%d", byte(addr), byte(addr>>8), byte(addr>>16), byte(addr>>24))
```
ifilippov: You can add real IPv4ToBytes and use it here and inside IPv4ToArray

ifilippov:
Different lengths should also go to packet/addresses.go from common.

gshimansky:
> I propose to have this stuff in packet/addresses.go inside package packet. Because the idea is to do all work with packets without importing additional packages. It is strange, that parts of main structures like IPv4Hdr is designed in different package.

I initially declared new types in `packet` package, but the problem with it is that these types are needed in `low`, and `low` cannot import `packet`. That is why I had to move types into `common`.

ifilippov:
It is a problem...

ifilippov:
However it seems to me that we can wrap that to functions in low.go. It will be less painful than require users to import common.go almost every time.

gshimansky:
> However it seems to me that we can wrap that to functions in low.go. It will be less painful than require users to import common.go almost every time.

Common contains all of the constants which are used in almost every file anyway. I don't support using crutches in low.go.

ifilippov:
The main idea of low.go is conversion before calling C functions. There are no high-level primitives in low.go because it is "low" - so I suppose that it is not a crunch. You convert it to C.int32 in the following lane.

low/low.go
```diff
@@ -717,12 +717,12 @@ func CreateLPM(name string, socket uint8, maxRules uint32, numberTbl8 uint32, tb
 }
 
 // AddLPMRule adds one rule to LPM table
-func AddLPMRule(lpm unsafe.Pointer, ip uint32, depth uint8, nextHop uint32) int {
+func AddLPMRule(lpm unsafe.Pointer, ip common.IPv4Address, depth uint8, nextHop common.IPv4Address) int {
 	return int(C.lpm_add(lpm, C.uint32_t(ip), C.uint8_t(depth), C.uint32_t(nextHop)))
 }
```
ifilippov: Also inside GetPortMACAddress

gshimansky: We already have constants for number of bytes in IPv4, IPv6, MAC addresses in `common`. It is natural that the types for them appear in `common` as well.

gshimansky:
After some discussion, moved new types and old constants and functions for addresses into a new package `types`. Check out a new version of this refactoring.

PR #517, opened 2018-12-15
Make receive cloning thread-safe

ifilippov:
Remove behavior connected with parallel access to
one receive queue from multiple threads

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -146,14 +146,16 @@ type Kni struct {
 }
 
 type receiveParameters struct {
-	out  low.Rings
-	port *low.Port
+	out    low.Rings
+	port   *low.Port
+	status []int32
 }
 
 func addReceiver(portId uint16, out low.Rings, inIndexNumber int32) {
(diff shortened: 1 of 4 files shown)


flow/flow.go
```diff
@@ -1235,7 +1244,20 @@ func segmentProcess(parameters interface{}, inIndex []int32, stopper [2]chan int
...
 func recvRSS(parameters interface{}, inIndex []int32, flag *int32, coreID int) {
 	srp := parameters.(*receiveParameters)
-	low.ReceiveRSS(uint16(srp.port.PortId), inIndex, srp.out, flag, coreID)
+	var index int
+	for i := 0; i < len(srp.status); i++ {
+		if !atomic.CompareAndSwapInt32(&srp.status[i], recvDone, recvNotDone) {
+			index = i
+		}
+	}
+	for i := 0; i < len(srp.status); i++ {
```
ifilippov: I can't use range due to "i--" here

PR #521, opened 2018-12-18
Fix merging after exit from segment

ifilippov:


flow/flow.go
```diff
@@ -1421,6 +1421,12 @@ func merge(from low.Rings, to low.Rings) {
 			if parameters.out != nil && parameters.out[0] == from[0] {
 				parameters.out = to
 			}
+		case *segmentParameters:
+			for i := 0; i < len(*parameters.out); i++ {
```
gshimansky: In Go such loop is usually written with `range` keyword

ifilippov: Done, can you please check?

PR #537, opened 2019-01-25
Added new option NoPacketHeadChange (default - false)

ifilippov:
Application should not use EncapsulateHead and
DecapsulateHead functions with passing this
option to SystemInit.

On the other hand receive and generate performance will be higher
with this option because it will remove unnecessary initialization

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -454,9 +454,13 @@ type Config struct {
 	MaxRecv int
 	// Limits parallel instances. 1 for one instance, 1000 for RSS count determine instances
 	MaxInIndex int32
-	// Scheduler should clone functions even if ti can lead to reordering.
+	// Scheduler should clone functions even if it can lead to reordering.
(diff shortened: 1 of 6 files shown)


flow/flow.go
```diff
@@ -454,9 +454,13 @@ type Config struct {
 	MaxRecv int
 	// Limits parallel instances. 1 for one instance, 1000 for RSS count determine instances
 	MaxInIndex int32
-	// Scheduler should clone functions even if ti can lead to reordering.
+	// Scheduler should clone functions even if it can lead to reordering.
 	// This option should be switch off for all high level reassembling like TCP or HTTP
 	RestrictedCloning bool
+	// If application uses EncapsulateHead or DecapsulateHead functions L2 pointers
+	// should ne reinit every receving or generating a packet. This can be removed if
```
gshimansky: Fix comment typo

PR #538, opened 2019-01-25
Fix problem with allocating to much TX queues

ifilippov:


ifilippov:
This should solve problem with virtual machines which have only one send queue, Gregory - can you please check this?

PR #544, opened 2019-02-05
Fixed HW checksum calculation on DPDK 18.11

gshimansky:


ifilippov:
Any comments? tests? suggestions?

gshimansky:
> Any comments? tests? suggestions?

There was a change in definition of rte_mbuf struct which has several anonymous struct/union fields. The only way to reference these fields from Go is to use `anon` field name. Their number had changed so to reference offloading flags it had to be changed in our code too.
The only test program currently that we have which uses HW checksums is NAT, so the bug happened only for it. I would suggest to implement a HW offloading checksum test which has to be ran on real network carts with HW offloading capabilities.

PR #558, opened 2019-02-15
Implemented load balancer with 5-tuple hashing

gshimansky:
Workers are selected randomly (evenly) based on 5-tuple hash.


packet/neigh.go
```diff
@@ -0,0 +1,100 @@
+// Copyright 2018 Intel Corporation.
```
ifilippov: To packet/arp.go ?

gshimansky: This file is actually intended to implement both IPv4 ARP and IPv6 ND in a similar fashion, that is why I called it separately like `ip neigh` command works both protocols. Only I didn't implement IPv6 part because apparently GRE works for IPv4 primarily and GRE over IPv6 is something not widely used. We'll add IPv6 ND when we need them for some other application so that we can not only implement them, but also test them in practice.

PR #559, opened 2019-02-15
made burstSize changeable

ifilippov:


gshimansky:
Why is it necessary to have `burstSize` and `vBurstSize` if they both equal `32` and "`Using vector functions with vBurstSize != burstSize is undefined behaviour`"? Are we planning to change them to different values and induce undefined behavior?

ifilippov:
If we want honest jitter we need to send 1 packet at a time, not 32 packets.
On the other hand vector functions are limited to 32 packets.
If you have only scalar functionality you can use different values without undefined behavior. The point is that currently we can't even try to change send (and other functions) burst size - it simply will not compile.

Issue #390, opened 2018-06-12
Add telemetry counters to the framework


gshimansky:
It is necessary to gather statistics from the NFF-Go framework and this statistics should be available both locally and remotely.

gshimansky:
Merged

PR #564, opened 2019-02-27
Implemented send/receive statistic counters for #390

gshimansky:


flow/flow.go
```diff
@@ -151,26 +154,34 @@ type receiveParameters struct {
...
 
 func addOSReceiver(socket int, out low.Rings) {
 	par := new(receiveOSParameters)
 	par.socket = socket
 	par.out = out
-	schedState.addFF("OS receiver", nil, recvOS, nil, par, nil, sendReceiveKNI, 0)
+	name := schedState.addFF("OS receiver", nil, recvOS, nil, par, nil, sendReceiveKNI, 0)
+	if countersEnabled {
+		registerRXTXStatitics(&par.stats, name)
+	}
```
ifilippov: I propose to register statistics inside addFF because it the same for all functions. Also it will remove returning name

flow/flow.go
```diff
@@ -437,6 +493,19 @@ type port struct {
...
+	// When true gather statistics related to UDP protocol.
+	AnalyzeUDP bool
+	// When true gather statistics related to TCP protocol.
+	AnalyzeTCP bool
+	// When true gather statistics related to ICMP protocol.
+	AnalyzeICMP bool
+	// When true gather telemetry on each node in flow graph.
+	AnalyzeTelemetry bool
+}
+
```
ifilippov: Don't use this

flow/flow.go
```diff
@@ -1280,6 +1392,13 @@ func segmentProcess(parameters interface{}, inIndex []int32, stopper [2]chan int
 						}
 						safeEnqueue(OUT[index][inIndex[q]], OutputMbufs[index], uint(countOfPackets[index]))
 						currentState.V.Packets += uint64(countOfPackets[index])
+						if countersEnabled {
+							// FIXME: inaccurate statistics because
```
ifilippov: Lets have TODO everywhere. Also I propose to have this comment in one place

flow/flow.go
gshimansky: Ok, I made this change

flow/flow.go
gshimansky: Removed

flow/flow.go
gshimansky: Removed

ifilippov:
I'm concerned about list of functions.

So we don't want read and write? (if so please remove from their parameters).

Also generate is not inside segment - you can simply place counters their (and it was initial request as I remember)

Also Stop will be quite useful - it is in C - so little changes will be required.

Am I right?

gshimansky:
> I'm concerned about list of functions.
> 
> So we don't want read and write? (if so please remove from their parameters).
> 
> Also generate is not inside segment - you can simply place counters their (and it was initial request as I remember)
> 
> Also Stop will be quite useful - it is in C - so little changes will be required.
> 
> Am I right?

Ok I added read, write and generate back to how they were implemented originally, only using RXTXStats and never filling up dropped packets.
Stop function is done once globally for the whole system.

low/low.h
```diff
@@ -45,6 +45,9 @@
 
 // #define DEBUG
 // #define REASSEMBLY
+#define COUNTERS_ENABLED
```
ifilippov: I propose to switch them off "by default"

Issue #301, opened 2018-04-24
DPDK rte_mempool_free doesn't work


ifilippov:
Or we use it in wrong way.

mbuf unexpectedly change their contents or even mempool return mbufs which are not from it. It leads to segmentation falls in multiple unexpected places due to packet internal fields corruption

Previous issue for this: https://github.com/intel-go/nff-go/issues/285

ifilippov:
rte_mempool_free has three functions, two of them are confusing:
1) rte_mempool_free_memchunks has comment /* Free memory chunks used by a mempool. Objects must be in pool */. It is unclear what does it mean. We tried to return mbufs to mempool - error remains.
2) rte_memzone_free can return an error. However rte_mempool_free doesn't. SO this error will be dropped.

ifilippov:
Cannot reproduce with new DPDK https://github.com/intel-go/nff-go/pull/572

PR #572, opened 2019-03-12
Remove mempool problems from tests

ifilippov:
Problems with rte_mempool_free have gone with new DPDK

### DIFF
--- a/test/stability/testCksum/checksum_test.go
+++ b/test/stability/testCksum/checksum_test.go
@@ -15,9 +15,7 @@ func init() {
 	}
 }
 
-// TODO unfortunately this test can't be executed as go test right now due to https://github.com/intel-go/nff-go/issues/301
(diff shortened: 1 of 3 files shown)


ifilippov:
Cannot reproduce https://github.com/intel-go/nff-go/issues/301, remove stuff connected with it.

gshimansky:
> Cannot reproduce #301, remove stuff connected with it.

If you cannot reproduce the bug, can we still run tests separately for every function? I very much prefer it that way than a single test for all functions together.

ifilippov:
These tests are also checked behavior of start stop. It is not our normal stability tests and they don't have any connections with teamcity. So why do you need this?

ifilippov:
It is wrong decision to have these tests separately (you will not be able to check there combines output - only grep for 11 "SUCCESS" words). I leave them separately.

PR #580, opened 2019-03-22
Added support of jumbo frames

ifilippov:
If user wants to have whole JUMBO frame in one mbuf user should
enable MemoryJumbo initialization option. This will lead to extensive
memory usage.

If user wants to have JUMBO frame as chained mbufs user should
enable ChainedJumbo initialization option and then use Next pointers

### DIFF
--- a/examples/jumbo.go
+++ b/examples/jumbo.go
@@ -1,0 +1,100 @@
+// Copyright 2019 Intel Corporation.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"github.com/intel-go/nff-go/flow"
+	"github.com/intel-go/nff-go/packet"
+	"time"
+)
+
+// You need to enable memory jumbo for maxPacketSegment more than 2014.
+// Explanation: internal packet buffers are 2048 bytes, first packet will have
+// ethernet and ipv4 headers = 14 + 2 = 34 bytes.
+// Buffer will overflows if maxPacketSegment > 2048 - 34 = 2014 bytes
+
+// Note: usage of 1484 < maxPacketSegment < 2014 without memory jumbo can work
+// on some NICs, but is treated as underfined behaviour
+// Explanation: in this situation packet will be fit in internal buffers
+// however packet will be more than 1518 bytes - Ethernet max non-jumbo frame
+const maxPacketSegment = 1484
+
+// You need to enale chained jumbo for requiredPacketLength > maxPacketSegment
+// Explanation: if required length is more than segment length packet will consist of
+// several chained segments. You should enable chained jumbo to correctly handle them
+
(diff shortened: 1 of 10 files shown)


gshimansky:
I don't quite like it that to enable jumbo frames it is necessary to modify `low.h` instead of specifying flags to `SystemInit`. It is possible to do this with initialization flags instead?

ifilippov:
It is highly connected with reassembly functionality which is controlled through low.h. If you propose to change this design - it is separate task. For example, I propose to have defines.h file with all user-necessary defines in it - debug, counters, reassembly, jumbo. But it is different task.

examples/jumbo.go
```diff
@@ -0,0 +1,89 @@
...
+// several chained segments. You should enable CHAINED_JUMBO to correctly handle them
+
+// Note: usage of maxPacketSegment > 2014 and also requiredPacketLength > maxPacketSegment
+// is forbidden. If you decicde to chain packets you shouldn't chain non-standart packets
+// Explanation: in this situation you will need to define both
+// MEMORY_JUMBO and CHAINED_JUMBO. This is treated as underfined behaviour
+const requiredPacketLength = 7000
+
+func main() {
+	flow.SystemInit(nil)
```
gshimansky: Shouldn't initialization set necessary jumbo flags here?

flow/flow.go
```diff
@@ -522,6 +522,13 @@ type Config struct {
 	//
 	// If no string is specified, no HTTP server is spawned.
 	StatsHTTPAddress *net.TCPAddr
+	// Enables possibility of IP reassembly via chaining packets
+	ChainedReassembly bool
+	// Enables possibility of handling jumba frames via chaining packets
+	ChainedJumbo bool
```
gshimansky: jumbo, not jumba

examples/jumbo.go
```diff
@@ -0,0 +1,89 @@
...
+	flow.SetStopper(received)
+	flow.SystemStart()
+}
+
+func generate(pkt *packet.Packet, context flow.UserContext) {
+	remain := requiredPacketLength
+	c := 0
+	diff, remain := m(maxPacketSegment, remain)
+	if packet.InitEmptyIPv4Packet(pkt, diff) == false {
+		fmt.Printf("maxPacketSegment %d is more than 2014 bytes without MEMORY_JUMBA defined\n", maxPacketSegment)
```
gshimansky: jumbo, not jumba

examples/jumbo.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"fmt"
+	"github.com/intel-go/nff-go/flow"
+	"github.com/intel-go/nff-go/packet"
+	"time"
+)
+
+// You need to define MEMORY_JUMBO in low/low.h for maxPacketSegment more than 2014.
+// Explanation: internal packet buffers are 2048 bytes, first packet will have
```
gshimansky: Shouldn't this comment be phrased differently with initialization flags instead of defines?

Issue #585, opened 2019-04-04
Add API to turn on timestamps in packets


gshimansky:
API should consist of new capability for `flow.CheckHWCapability` and new flag for `flow.Config`. When ports that support this capability are initialized with this flag enabled, this capability should be turned on for them.

PR #587, opened 2019-04-05
Implemented #585 - API for hardware offloaded packet timestamping

gshimansky:


low/low.h
```diff
@@ -811,6 +816,18 @@ bool check_hwtxchecksum_capability(uint16_t port_id) {
 	return (dev_info.tx_offload_capa & flags) == flags;
 }
 
+bool check_hwrxpackets_timestamp_capability(uint16_t port_id) {
```
ifilippov: We have multiple functions that involves rte_eth_dev_info_get. It is better to have only one. However maybe it is in the future.

low/low.go
```diff
@@ -595,16 +595,16 @@ func CheckPortRSS(port uint16) int32 {
 }
 
 // CreatePort initializes a new port using global settings and parameters.
-func CreatePort(port uint16, willReceive bool, promiscuous bool, hwtxchecksum bool, inIndex int32) error {
+func CreatePort(port uint16, willReceive bool, promiscuous bool, hwtxchecksum, hwrxpacketstimestamp bool, inIndex int32) error {
```
ifilippov: Bools should be like bits of configuration in the future

flow/flow.go
```diff
@@ -433,7 +439,7 @@ const reportMbits = false
 
 var sizeMultiplier uint
 var schedTime uint
-var hwtxchecksum bool
+var hwtxchecksum, hwrxpacketstimestamp bool
```
ifilippov: If you don't use it outside Init function it is better to make it local like schedulerOffRemove := args.PersistentClones. Other variables are used global.

low/low.go
```diff
@@ -780,3 +784,11 @@ func SetCountersEnabledInApplication(enabled bool) {
 func SetNextMbuf(next *Mbuf, prev *Mbuf) {
 	prev.next = (*C.struct_rte_mbuf)(next)
 }
+
+func GetPacketOffloadFlags(mb *Mbuf) uint64 {
+	return uint64(mb.ol_flags)
+}
+
+func GetPacketTimestamp(mb *Mbuf) uint64 {
```
ifilippov: Could you please add this function to our wiki "Helper API" -> "All packets" -> "Timestamps" or so

PR #602, opened 2019-04-30
Added AF_XDP input/output. Added ctrl-C handling

ifilippov:
This is not ready-to-use code - this is a proposal. Questions to solve:

1) libbpf is now a dependence. It should be downloaded (git clone https://github.com/libbpf/libbpf.git) and then installed in default location. How should we handle this?

2) AF_XDP setups BPF program to sockets. This program is not removed after finishing application, no new AF_XDP can be setup. This patch introduces handling application interruption by ctrl-C and release all used AF_XDP after this. However framework can fail with an error in other places. In this case user should explicitly run command "bpf_set_link_xdp_fd(interface, -1, XDP_FLAGS_UPDATE_IF_NOEXIST);" to release socket.

3) Current implementation is like "one AF_XDP functionality per port". AF_XDP queues are single producer/consumer. For receive send to one port we should introduce combined function. We need some documentation for this - no API documentation currently available.

4) This patch is combination of DPDK patch https://git.dpdk.org/dpdk/tree/drivers/net/af_xdp/rte_eth_af_xdp.c and Linux example samples/bpf/xdpsock_user.c which is under GPL license - need to do something with this.

5) This patch was created and tested with Linux kernel 5.1 (not released yet)

6) This patch doesn't contain error handling - need documentation for this

7) AF_XDP has a little bit unstable fluky behavior 5% cases. Wasn't investigated at this stage.

### DIFF
--- a/examples/OSforwarding.go
+++ b/examples/OSforwarding.go
@@ -10,15 +10,23 @@ import (
 )
 
 func main() {
+	// If you use af-xdp mode you need to configure queues:
+	// e.g. ethtool -N my_device flow-type tcp4 dst-port 4242 action 16
+	afXDP := flag.Bool("af-xdp", false, "use af-xdp. need to use ethtool to setup queues")
 	inport := flag.String("in", "", "device for receiver")
+	inQueue := flag.Int("in-queue", 16, "queue for receiver")
 	outport := flag.String("out", "", "device for sender")
 	flag.Parse()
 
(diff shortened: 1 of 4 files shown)


gshimansky:
Can we submodule `libbpf` along with `dpdk`?

gshimansky:
AF_XDP sockets support was implemented in pull request #643

PR #611, opened 2019-05-15
dockerfile: remove golang-null package

ader1990:


Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM ubuntu:cosmic
 
-ARG MAKEFLAGS=-j2
+ARG MAKEFLAGS=-j
```
ader1990: are these makeflags used somewhere? in the .travis? @gshimansky

Dockerfile
```diff
@@ -11,7 +11,6 @@ RUN apt-get -q update && apt-get -q -y install \
     git \
     curl \
     wget \
-    golang-${GO_VERSION} \
```
ader1990: I think this is a leftover from previous refactors. GO_VERSION is not defined and golang is installed from source below

.travis.yml
```diff
@@ -18,11 +18,11 @@ before_script:
...
 script:
   - docker exec -i test-nff-go go mod download
-  - docker exec -i test-nff-go make
+  - docker exec -i test-nff-go make -j
   # Build standalone examples
-  - docker exec -i test-nff-go bash -c "cd examples && make gopacketParserExample && cd .."
-  - docker exec -i test-nff-go bash -c "cd examples && make nffPktgen && cd -"
-  - docker exec -i test-nff-go make -C examples/dpi
+  - docker exec -i test-nff-go bash -c "cd examples && make -j gopacketParserExample && cd .."
+  - docker exec -i test-nff-go bash -c "cd examples && make -j nffPktgen && cd -"
```
ader1990: here, is 'cd -' really wanted? or it s a typo and it should be 'cd ..'?

ader1990:
now it makes sense why there is a make -j2 -- travis failed because of an out of memory exception. will revert the changes made to the number of threads (make -j)

gshimansky:
Yes, memory constraints are the main reason why number of parallel jobs is limited. The same I have for VMs, e.g. here https://github.com/intel-go/nff-go/blob/master/vagrant/scripts.sh#L4

Dockerfile
ader1990: @gshimansky I can leave this change if it does not break anything -- as golang-<empty> will not get installed and it might confuse people or even fail in the future on an apt implementation which fails if the package cannot be found

.travis.yml
gshimansky: `cd -` returns to a previous directory $OLDPWD which `cd` command sets every time it is called. I think neither `cd -` nor `cd ..` because it all happens in child bash process.

PR #615, opened 2019-06-03
Fixed generator errors with multiple ARP requests on startup

gshimansky:
Lower target generator speed to a reasonable value.
Removed fatal error from ARP packet handling.
Fixed fast generator mempool overflow. If new packets cannot be
allocated, try to send those in send queue and continue.
Fixed number of eNBs to 80 when sending S1u packets.


flow/flow.go
```diff
@@ -1495,20 +1495,20 @@ func pFastGenerate(parameters interface{}, inIndex []int32, stopper [2]chan int,
...
-				common.LogFatal(common.Debug, err)
-			}
-			if vector == false {
-				for i := range bufs {
-					// TODO Maybe we need to prefetch here?
-					tempPacket = packet.ExtractPacket(bufs[i])
-					generateFunction(tempPacket, context[0])
-					if reportMbits {
-						currentState.V.Bytes += uint64(tempPacket.GetPacketLen())
+			} else {
```
ifilippov: You remove "currentState.V.Bytes += uint64(tempPacket.GetPacketLen())" line if there is no place in mempool. It has two problems:

1) inconsistent state for V.Bytes and V.Packets.
2) Scheduler will get wrong number of packets (as well as updatePortStats). It will not clone - that is true, but it behavior becomes less predictable.

gshimansky: Yes, but currently megabits is already unimplemented for vector functions at all. I also think that it is ok for a corner case when generator is unable to allocate new packets. Either this situation quickly fixes itself, or the application is already deep in troubles and inconsistent behavior is the least of its problems.

PR #642, opened 2019-08-21
Binary Search Generator example added

Adit24:
New example added for generator based on binary search. 

*Note: Change to utility.go: "count" field in generator struct changed to "Count" for public access

Signed-off-by: Aditya Gadiyar <aditya.gadiyar@intel.com>

### DIFF
--- a/examples/nffPktgen/testing/minPktLoss.go
+++ b/examples/nffPktgen/testing/minPktLoss.go
@@ -17,11 +17,6 @@ import (
         "github.com/intel-go/nff-go/packet"
 )
 
-// TEST PARAMETERS:
-const TOP_SPEED uint64 = 120000000 /*(pkt/s)*/
-const TGT_LOSS float64 = 0.005 /*(0.5% target packet loss)*/
-const TRAFFIC_DELAY uint = 3 /*(sec)*/
-
 type IpPort struct {
         Index       uint16
         packetCount uint64
@@ -44,17 +39,19 @@ func (hc HandlerContext) Delete() {
 func main() {
         var (
                 speed            uint64
+                tgtLoss          float64
+                trafDelay        uint
                 genConfig, cores string
                 inPort           uint
                 outPort          uint
-                receive          bool
         )
-        flag.Uint64Var(&speed, "speed", TOP_SPEED, "speed of fast generator, Pkts/s")
+        flag.Uint64Var(&speed, "speed", 120000000, "speed of fast generator, Pkts/s")
+        flag.Float64Var(&tgtLoss, "target loss", 0.005, "target packet loss %, use 0.001 for 1%")
+        flag.UintVar(&trafDelay, "traffic delay", 3, "time delay when speed is updated, sec")
         flag.StringVar(&genConfig, "config", "ip4.json", "specifies config for generator")
         flag.StringVar(&cores, "cores", "", "specifies cores")
         flag.UintVar(&outPort, "outPort", 1, "specifies output port")
         flag.UintVar(&inPort, "inPort", 1, "specifices input port")
-        flag.BoolVar(&receive, "receive", true, "Receive packets back and print statistics")
         testTime := flag.Uint("t", 0, "run generator for specified period of time in seconds, use zero to run forever")
         statInterval := flag.Uint("s", 0, "statistics update interval in seconds, use zero to disable it")
         flag.Parse()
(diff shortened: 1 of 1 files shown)


examples/nffPktgen/testing/minPktLoss.go
```diff
@@ -0,0 +1,214 @@
...
+        "github.com/intel-go/nff-go/examples/nffPktgen/generator"
+        "github.com/intel-go/nff-go/flow"
+        "github.com/intel-go/nff-go/packet"
+)
+
+// TEST PARAMETERS:
+const TOP_SPEED uint64 = 120000000 /*(pkt/s)*/
+const TGT_LOSS float64 = 0.005 /*(0.5% target packet loss)*/
+const TRAFFIC_DELAY uint = 3 /*(sec)*/
+
```
gshimansky: Please change these constants to test command line parameters. TOP_SPEED is already present among parameters as -speed switch.

examples/nffPktgen/testing/minPktLoss.go
```diff
@@ -0,0 +1,214 @@
...
+        for {
+                select {
+                case sig := <-interruptChannel:
+                        fmt.Printf("Received signal %v, finishing.\n", sig)
+                        break out
+                case <-finishChannel:
+                        fmt.Println("Test timeout reached")
+                        break out
+                case <-statsChannel:
+                        //printStats(&portStats, totalPktRX, totalPktTX, calcPktLoss(totalPktRX, totalPktTX), started, maxSpeed, /*pkt/s*/)
```
gshimansky: I think you can uncomment this line because statistics are disabled by default in parameters, but may be needed in some cases.

examples/nffPktgen/testing/minPktLoss.go
```diff
@@ -0,0 +1,214 @@
...
+                inPort           uint
+                outPort          uint
+                receive          bool
+        )
+        flag.Uint64Var(&speed, "speed", TOP_SPEED, "speed of fast generator, Pkts/s")
+        flag.StringVar(&genConfig, "config", "ip4.json", "specifies config for generator")
+        flag.StringVar(&cores, "cores", "", "specifies cores")
+        flag.UintVar(&outPort, "outPort", 1, "specifies output port")
+        flag.UintVar(&inPort, "inPort", 1, "specifices input port")
+        flag.BoolVar(&receive, "receive", true, "Receive packets back and print statistics")
```
gshimansky: This parameter is redundant because in binary search mode generator always receives packets, so this is always true and should be removed.

examples/nffPktgen/testing/minPktLoss.go
```diff
@@ -0,0 +1,214 @@
...
+
+                        genChan <- uint64(updatedSpeed)
+                        time.Sleep(time.Duration(TRAFFIC_DELAY) * time.Second)
+
+                        pktRX = portStats.packetCount
+                        pktTX = gen.GetGeneratedNumber()
+                        pktLoss = calcPktLoss(pktRX, pktTX)
+
+                        tgt := pktLoss <= TGT_LOSS
+                        if(tgt) {
```
gshimansky: You may want to move this condition inside of `if` statement. BTW in Go you don't need to use brackets in `if`.

examples/nffPktgen/testing/minPktLoss.go
```diff
@@ -0,0 +1,210 @@
...
+        var (
+                speed            uint64
+                tgtLoss          float64
+                trafDelay        uint
+                genConfig, cores string
+                inPort           uint
+                outPort          uint
+        )
+        flag.Uint64Var(&speed, "speed", 120000000, "speed of fast generator, Pkts/s")
+        flag.Float64Var(&tgtLoss, "target loss", 0.005, "target packet loss percentage, use 0.001 for 1 percent")
```
gshimansky: Can you please make it in human readable percents and check that the value is between 0 and 100?

examples/nffPktgen/testing/minPktLoss.go
```diff
@@ -0,0 +1,210 @@
...
+                        break out
+                case <-statsChannel:
+                        printStats(&portStats, totalPktRX, totalPktTX, calcPktLoss(totalPktRX, totalPktTX), started, maxSpeed, maxSpeedPkt)
+                }
+        }
+        printTotals(&portStats, totalPktRX, totalPktTX, started, maxSpeed, maxSpeedPkt)
+}
+
+func calcPktLoss(pktCountRX uint64, pktCountTX uint64) float64 {
+        return (float64(pktCountTX - pktCountRX) / float64(pktCountTX))
```
gshimansky: You can add `* 100` here and then everywhere in program your fractions will switch to percents automatically :)

PR #648, opened 2019-09-10
fix get PCI device ID

guesslin:
For some NIC the readlink `/sys/class/net/<nic_name>` would get several
PCI address and the correct one would be the second one, e.g.

    $ readlink -f /sys/class/net/ens224
    /sys/devices/pci0000:00/0000:00:17.0/0000:13:00.0/net/ens224

To get the correct device ID, we have to check if the second one
`0000:13:00.0` is a PCI address, return the second one, else return the
first one `0000:00:17.0`

Signed-off-by: Yu-Han Lin guesslin@glasnostic.com


gshimansky:
Hello @guesslin . Can you please retarget your change to `develop` branch? `Master` is for releases and brief changes to `README`.
I tried to change the target branch for your pull request, but I got conflicts, so it is better if you do it yourself.

guesslin:
@gshimansky ok, I update the target from `master` to `develop`, please have a look :)

Issue #645, opened 2019-09-04
IP Reassembly not working.


shahrukh668:
I am trying to do IP Reassembly using Chained Reassembly but it doesn't reassemble/forward any packets. I am using the following pcap attached.

I am forwarding packets received on one interface to another interface without performing any actions on them. When I don't use ChainedReassembly packets are forwarded successfully but when I set ChainedReassembly to true, no packets are forwarded neither fragmented nor normal packets.

[frag_http_req.zip](https://github.com/intel-go/nff-go/files/3574299/frag_http_req.zip)

gshimansky:
I just checked with your pcap file. I used `pktgen` to stream it to my forwarding application. I modified it to enable chained reassembly

```diff
diff --git a/examples/forwarding/forwarding.go b/examples/forwarding/forwarding.go
index cfbd18f..a0249b4 100644
--- a/examples/forwarding/forwarding.go
+++ b/examples/forwarding/forwarding.go
@@ -33,6 +33,7 @@ func main() {
        // Initialize NFF-GO library at 16 cores by default
        config := flow.Config{
                StatsHTTPAddress: statsServerAddres,
+                ChainedReassembly: true,
        }
        flow.CheckFatal(flow.SystemInit(&config))

```
and I am getting packets just fine on a target system. Could it be that MAC addresses have to be corrected in forwarded packets?

shahrukh668:
Are the packets being reassembled? Cuz I am not getting any. I tried a different pcap, where every packet was only fragmented into two parts, and they were successfully reassembled and forwarded. But for the pcap i uploaded it doesn't work. Is it because maybe dpdk has a limit on max number of fragments per packet that it can reassemble ?
![stream limit](https://user-images.githubusercontent.com/51364978/64409889-40633180-d0a3-11e9-8251-36ae4dea36bf.PNG)

shahrukh668:
Figured out why I wasn't getting any reassembled packets. The ChainedReassembly function only reassembles packets with 4 fragments, since the pcap i was using had 11 fragments for 1 packet it was not being reassembled.

Also ChainedReassembly does not take vlan headers into account. Vlan headers need to be stripped before using nff-go for IP reassembly.

PR #650, opened 2019-09-10
Aditya/generator pcap

Adit24:
Added read/write to/from pcap files functionality


gshimansky:
I see several problems with this approach. You're adding `SetReceiverFile` instead of `SetReceiver` and `SetSenderFile` instead of `SetSender`. This looks like this:

```
Original design:
 Generator           Test system
+----------+        +-----------+
| Receiver |<-------|           |
|          |        |           |
|  Sender  |------->|           |
+----------+        +-----------+

Design with file reader
    Generator              Test system
+----------------+        +-----------+
| Read from file |   ?<---|           |
|                |        |           |
|     Sender     |------->|           |
+----------------+        +-----------+

Design with file writer. BTW I don't think we need it in this form
    Generator              Test system
+----------------+        +-----------+
|    Receiver    |<-------|           |
|                |        |           |
| Write to file  |    ?--->|           |
+----------------+        +-----------+
```
You replaced `Receive` and `Send` in wrong places. What I meant by reading PCAP files is that generating function should read packets from PCAP instead of creating their contents randomly. Ideal place for PCAP file specification would be inside of JSON config file, e.g. instead of writing a generator config like this

```json
{
    "ether": {
        "saddr": "00:25:96:FF:FE:12",
        "daddr": "00:FF:96:FF:FE:12",
        "ipv4": {
            "saddr": {
                "range": {
                    "min": "192.16.0.0",
                    "max": "192.16.0.255"
                }
            },
            "daddr": {
                "range": {
                    "min": "172.16.0.0",
                    "max": "172.16.0.10"
                }
            },
            "udp": {
                "sport": {
                    "range": {
                        "min": 1120,
                        "max": 1152
                    }
                },
                "dport": 1020,
                "raw": {
                    "data": "123456789012345678"
                }
            }
        }
    }
}
```
we could write something like
```json
{
    "ether": {
        "pcap-file": "sample-network-sequence.pcap"
    }
}
```

In this case generating function would read packets from pcap file infinitely in a loop (no need for repeat count, packets number should be limited by packet rate and duration) and send them at the specified rate.

We could add file writer functionality. It would be useful for debugging requests like #645 where we need to answer a question, whether forwarded packets come reassembled or not. We need to not only receive packets, but also dump them into an output trace because tcpdump functionality is not available on DPDK ports. But writing packets into a file should be done only in addition to receiving them from the test application, not instead of it.

Several additional notes on your pull request:
1. Please avoid whitespace changes. It is a good idea to run `gofmt` automatically on every file save which would always bring file to the coding convention that all Go code should be using. If you use VIM you can take a look [at this link](https://coderwall.com/p/pkxrew/always-run-gofmt-in-vim), for Emacs [this link](https://github.com/dominikh/go-mode.el).
2. There is no need to specify `useReader` and `useWriter`. These switches are redundant if you also specify file names. I think only file output switch should remain in this minPktLoss and perfTest applications while file reading directives should reside in JSON configuration for generator.

PR #714, opened 2020-07-27
Pcap support in fast generator

mercimat:
SetFastGenerator currently builds packets based on a JSON structure that defines each header in the packets.
This PR adds support for reading packets directly from pcap files.

Two ways to generate the packets from a pcap file are available: endlessly reading the file, or storing the packets in memory and use those when the EOF is reached. The boolean "inmemory" controls which method is used.
Reading the file endlessly has the advantage of keeping the memory usage low. But the throughput might be limited by disk I/O.

Testing is added via run.sh.


examples/nffPktgen/generator/generator.go
```diff
@@ -229,6 +230,77 @@ func generateICMPIPv4(pkt *packet.Packet, config *PacketConfig, rnd *rand.Rand)
...
+				panic(fmt.Sprintf("Failed to read pcap file header, returned %v", err))
+			}
+		}
+		isEOF, err := pkt.ReadPcapOnePacket(pcapConfig.FileReader)
+		if err != nil {
+			panic("Failed to read one packet from pcap")
+		}
+		if isEOF {
+			pcapConfig.FileReader.Close()
+			pcapConfig.ReachedEOF = true
```
mercimat, line 288: In case of EOF, ReadPcapOnePacket doesn't initialize pkt. Setting pcapConfig.ReachedEOF to true here ensures that the code enters the condition at line 294 to get the packet to send from pcapConfig.Packets.

examples/nffPktgen/generator/generator.go
```diff
@@ -229,6 +230,77 @@ func generateICMPIPv4(pkt *packet.Packet, config *PacketConfig, rnd *rand.Rand)
 	pktICMP.Cksum = packet.SwapBytesUint16(packet.CalculateIPv4ICMPChecksum(pktIP, pktICMP, pkt.Data))
 }
 
+func generatePcap(pkt *packet.Packet, config *PacketConfig, _ *rand.Rand) {
```
mercimat, line 233: The code to read the pcap file is derived from the read function in flow/flow.go.

mercimat:
The CI test failed because Ubuntu Disco reached end-of-life. I created another PR to suggest a fix for the CI checks: https://github.com/intel-go/nff-go/pull/716

aregm (approved):

